Issue #19, opened 2025-01-12
[Enhancement]: Add option to configure port and address


Norbiros:
### My idea

Now we start the server on `127.0.0.1:8080`, add option to modify this default behaviour using environment variables.

PR #21, opened 2025-01-12
enhancement: Enabled changing backend ip by env

Kubaryt:
Resolves: #19

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,2 +1,2 @@
 DATABASE_URL=postgres://backend:password@localhost:5432/hack4krak
-BACKEND_IP=127.0.0.1:8080
+BACKEND_ADDRESS=127.0.0.1:8080
--- a/.env.example
+++ b/.env.example
@@ -1,1 +1,2 @@
 DATABASE_URL=postgres://backend:password@localhost:5432/hack4krak
+BACKEND_IP=127.0.0.1:8080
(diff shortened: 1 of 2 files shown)


backend/src/main.rs
```diff
@@ -24,6 +24,12 @@ async fn main() -> std::io::Result<()> {
     let db: DatabaseConnection = Database::connect(database_url).await.unwrap();
     Migrator::up(&db, None).await.unwrap();
 
+    let ip_env = std::env::var("BACKEND_IP").unwrap();
```
Norbiros, line 27: Provide default value `127.0.0.1:8080`

.env.example
```diff
@@ -1 +1,2 @@
-DATABASE_URL=postgres://backend:password@localhost:5432/hack4krak
\ No newline at end of file
+DATABASE_URL=postgres://backend:password@localhost:5432/hack4krak
+BACKEND_IP=127.0.0.1:8080
```
Norbiros, line 2:
```suggestion
BACKEND_ADDRESS=127.0.0.1:8080
```

backend/src/main.rs
```diff
@@ -24,6 +24,12 @@ async fn main() -> std::io::Result<()> {
...
+    let mut ip_addr: Vec<&str> = vec![];
+    for addr_part in ip_env.split(":") {
+        ip_addr.push(addr_part);
+    }
```
Norbiros, line 31:
```suggestion
    let address_env = std::env::var("BACKEND_IP").unwrap().split(":").collect();
    let ip = adress_env[0];
    let port = adress_env[1].parse::<u16>().expect("The port in BACKEND_IP must be a valid u16 integer")
```

Norbiros (changes requested):

Norbiros (approved):

PR #30, opened 2025-01-15
fix: Ignore openapi json in eslint

Kubaryt:


Norbiros (changes requested):
Use `cargo run` and update `openapi.json` to not be formatted

Norbiros (approved):

Delop25 (approved):

Issue #75, opened 2025-01-27
[Bug]: Github OAuth fails for non-public emails


Norbiros:
### Description

When you have private emails on github and you try to sign in with github oauth you get 401 invalid credentials error, after further investigation I have found out that in response body email is null

### Steps to reproduce

1. Set your email in github account to private
2. Start backend server and go to the /auth/oauth/github endpoint
3. Error 401 wrong password and/or username

### Demonstration(s) and additional context

It's pretty simple fix: https://stackoverflow.com/questions/35373995/github-user-email-is-null-despite-useremail-scope

PR #76, opened 2025-01-27
fix: Fixed email being null in github oauth

Kubaryt:
Resolves: #75

### DIFF
--- a/backend/src/routes/auth/oauth.rs
+++ b/backend/src/routes/auth/oauth.rs
@@ -2,6 +2,7 @@ use crate::models::entities::users;
 use crate::routes::auth::TokensResponse;
 use crate::utils::app_state::AppState;
 use crate::utils::error::Error;
+use crate::utils::error::Error::InvalidCredentials;
 use actix_web::{get, web, HttpResponse};
 use oauth2::reqwest;
 use oauth2::{AuthorizationCode, CsrfToken, Scope, TokenResponse};
@@ -16,7 +17,13 @@ struct QueryParams {
 #[derive(Deserialize, Debug)]
 struct GitHubUser {
     name: String,
+    email: Option<String>,
+}
+
+#[derive(Deserialize, Debug)]
+struct GitHubEmail {
     email: String,
+    primary: bool,
 }
 
 #[utoipa::path(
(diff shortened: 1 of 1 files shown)


backend/src/routes/auth/oauth.rs
```diff
@@ -46,23 +53,41 @@ pub async fn github_callback(
...
-    if !response.status().is_success() {
+    if !response_user.status().is_success() {
         return Err(Error::InvalidCredentials);
     }
 
-    let user: GitHubUser = response
-        .json()
-        .await
-        .map_err(|_| Error::InvalidCredentials)?;
+    let mut user: GitHubUser = response_user.json().await.map_err(|_| InvalidCredentials)?;
```
Norbiros, line 67: Return different error, it will be eaisier to find those bugs in future. We should return `500` or something like that

backend/src/routes/auth/oauth.rs
```diff
@@ -46,23 +53,41 @@ pub async fn github_callback(
...
+            .await?
+            .json()
+            .await?;
+
+        if let Some(primary_email) = email_response.iter().find(|email| email.primary) {
+            user.email = Some(primary_email.email.clone());
+        }
+    }
+
+    if user.email.is_none() {
```
Norbiros, line 84: Use
```
let Some(email) = user.email else {
    return Err(InvalidCredentials);
}
```

backend/src/routes/auth/oauth.rs
```diff
@@ -46,23 +53,41 @@ pub async fn github_callback(
         .map_err(|_| Error::OAuth)?;
 
     let token = format!("Bearer {}", token_result.access_token().secret());
-    let response = reqwest::Client::new()
+    let response_user = reqwest::Client::new()
```
Norbiros, line 56: Duplicated code 🚨 🚨 🚨, create method for sending request to github api

Norbiros (changes requested):

backend/src/routes/auth/oauth.rs
```diff
@@ -16,7 +18,22 @@ struct QueryParams {
...
+    email: Option<String>,
+}
+
+#[derive(Deserialize, Debug)]
+struct GitHubEmail {
     email: String,
+    primary: bool,
+}
+
+pub async fn send_github_request(url: Url, token: &String) -> Result<Response, reqwest::Error> {
```
Norbiros, line 30:
```suggestion
async fn send_github_request(url: Url, token: &String) -> Result<Response, reqwest::Error> {
```

Norbiros (changes requested):
only that

Norbiros (approved):

PR #107, opened 2025-01-31
feat: Setup backend hosting

Norbiros:


Kubaryt (approved):
lgtm

backend/src/main.rs
```diff
@@ -102,11 +102,13 @@ async fn main() -> std::io::Result<()> {
             .openapi_service(|api| Scalar::with_url("/docs", api))
             .split_for_parts();
 
-        let path = &Config::get().openapi_json_frontend_path;
-        let mut openapi_json = File::create(path).unwrap();
-        openapi_json
-            .write_all(to_string(&api).unwrap().as_bytes())
-            .unwrap();
+        if cfg!(debug_assertions) {
```
wikipop, line 105: I dont have any f clue why this if exists (I asumme for tests to pass). Maybe put some comment on this?

wikipop (changes requested):

backend/src/main.rs
Norbiros: If you have questions about syntax/features of specific language, don't mark them as request changes.

Configuration conditional checks are used to test if some feature is enabled, here I want to generate `openapi_json` only if binary is built in debug mode, because there is no reason to generate file in `frontend` directory when running backend in production (release mode).

Norbiros: I even created separate commit for it:

> [fix(backend): Do not generate openapi schema on production](https://github.com/Hack4Krak/Hack4KrakSite/pull/107/commits/c94b7e465208c2711ed55a744384fba3339145c8)

wikipop: yeah mb I wasn't talking about my inability to read English, I was referring to the unreadability of your code

Norbiros: I think it is pretty straightforward. In debug mode, we enter the `if` statement and continue with execution. It's logical that we wouldn't want to generate the openapi.json in release mode.

Norbiros: @wikipop

wojpo (approved):

PR #112, opened 2025-02-01
[Fix/frontend]: Auth related fixes

Norbiros:


Kubaryt (approved):

wikipop (approved):
Ensure that no null data is populated into components on error

Norbiros:
> Ensure that no null data is populated into components on error

1. Not related to this pr
2. This depends on use case of `useAuth`, `useApi`, normal fetch etc.

wojpo (approved):

PR #117, opened 2025-02-02
feat: email integration test

wikipop:


backend/src/utils/app_state.rs
```diff
@@ -32,12 +32,15 @@ pub struct AppState {
 }
 
 impl AppState {
-    pub fn with_database(database: DatabaseConnection) -> AppState {
-        let oauth_client: OAuthClient = BasicClient::new(ClientId::new("test".to_string()))
+    fn default_oauth_factory() -> OAuthClient {
```
Norbiros, line 35: I don't like naming default constuctor `default_oauth_factory`. Just use `impl Default for AppState`

backend/tests/email.rs
```diff
@@ -0,0 +1,65 @@
...
+
+    let smtp_client = SmtpTransport::builder_dangerous(host.to_string())
+        .tls(Tls::None)
+        .port(smtp_port)
+        .build();
+
+    let email = Email {
+        sender: "user@domain.tld".to_string(),
+        subject: "Le Corbusier and his crimes".to_string(),
+        recipients: vec!["danger@domain.tld".to_string()],
```
Norbiros, line 34: Test multiple receipents

backend/tests/email.rs
```diff
@@ -0,0 +1,65 @@
...
+use hack4krak_backend::utils::emails::{Email, EmailTemplate};
+use lettre::transport::smtp::client::Tls;
+use lettre::SmtpTransport;
+use serde_json::Value;
+use testcontainers::core::{IntoContainerPort, WaitFor};
+use testcontainers::runners::AsyncRunner;
+use testcontainers::GenericImage;
+
+#[actix_web::test]
+async fn send_mail() {
```
Norbiros, line 12: Hmm... Maybe make tests using testcontainers opt-in? Some users might want to run other tests without running those.

backend/Cargo.toml
```diff
@@ -43,3 +43,7 @@ utoipa-scalar = { version = "0.3.0", features = ["actix-web"] }
 # Custom crates
 migration = { path = "migration" }
 regex = "1.11.1"
+
+[dev-dependencies]
+testcontainers = "0.23.1"
```
Norbiros, line 48: Mention it in `Backend/README.md`

backend/src/utils/error.rs
```diff
@@ -39,8 +39,10 @@ pub enum Error {
     InvalidJson(#[from] serde_json::Error),
     #[error("Email template not found")]
     EmailTemplateNotFound,
-    #[error("Invalid email address while sending email")]
-    InvalidEmailAddressSendingEmail,
+    #[error("Invalid sender's email")]
+    InvalidEmailSender,
+    #[error("Invalid recipients' email")]
+    InvalidEmailRecipients,
```
Norbiros, line 45: Add argument for this email, because I may have 5000 senders

backend/src/utils/error.rs
```diff
@@ -39,8 +39,10 @@ pub enum Error {
     InvalidJson(#[from] serde_json::Error),
     #[error("Email template not found")]
     EmailTemplateNotFound,
-    #[error("Invalid email address while sending email")]
-    InvalidEmailAddressSendingEmail,
+    #[error("Invalid sender's email")]
+    InvalidEmailSender,
```
Norbiros, line 43: Same as below

Norbiros (changes requested):

backend/src/utils/app_state.rs
wikipop: cant implement trait to `type`

backend/src/utils/error.rs
wikipop: how do u want have fokin many senders

backend/src/utils/app_state.rs
Norbiros: Whaaat? `pub struct AppState`

backend/src/utils/error.rs
Norbiros: For example:
- you send email to the entire team about disqualification
- we just have email queue
- or just you send emails one by one.

You want to know which mail was the "bad" one so you can just look to the logs and see what to fix

Kubaryt: iirc you can't have many **Senders** in letttre, at least I didn't implement it in Email struct

Norbiros: Oh senders, but it is pretty much the same. We may use one email for notifications (like ctf starts in 7 days), one for things related to auth and another for admin stuff, you want to easily know which one failed, because it will be easier to debug.

Kubaryt: but you don't use many adresses to send one email? because if you don't then I don't really know what you what to test in this case and if you do, then what for?

Norbiros: I think it is impossible to have multiple senders. But that isn't important. You want to store which email address was invalid inside your error, to make it easier to debug than just mysterious "invalid sender"

Kubaryt: oh i misunderstood, makes sense

backend/Cargo.toml
Norbiros: Inform about security, because most of people will be concerned why it asks for sudo to tests and mention better alternatives like podman. Also add note about the custom use flag for those sests

Norbiros (approved):

Norbiros (changes requested):
Just add new job to package.json inside backend

README.md
```diff
@@ -43,7 +43,11 @@ pnpm dev
 Because some backend tests are expensive and require docker (e.g. email integration tests), they are not run locally by default. To run the full test suite, use the following command:
 
 ```shell
-cargo test --features=full-test-suite
+# root dir
+pnpm backend:test:full
+
+# /backend dir
+pnpm test:full
```
Norbiros, line 50: Remove it. This is main `README.md`, not backend. Here we just mention global commands (so with `backend`)

Norbiros (changes requested):

README.md
wikipop: ik ik

Norbiros (approved):

Norbiros (approved):

backend/tests/email.rs
```diff
@@ -0,0 +1,83 @@
...
+        .port(smtp_port)
+        .build();
+
+    let email = Email {
+        sender: "user@domain.tld".to_string(),
+        subject: "Le Corbusier and his crimes".to_string(),
+        recipients: vec![
+            "danger@domain.tld".to_string(),
+            "dziegiel@domain.tld".to_string(),
+            "borbert@domain.tld".to_string(),
```
Kubaryt, line 39: change it to bobert

Kubaryt (approved):
lgtm

goteusz-maszyk (approved):

Issue #114, opened 2025-02-01
[Feat]: Setup user permissions


Norbiros:
### My suggestion

It would be nice if we could give roles to users, so in the future we could create admin dashboard etc. Create endpoints for changing role of specific user, checking etc.

Also create a middleware with option to set that this role is required.

And if you want, you can create /tasks/refresh endpoint

PR #135, opened 2025-02-03
Feat: Roles permission system

Kubaryt:
Resolves: #114


Norbiros (changes requested):
I think you rebased your branch incorrectly (it might have been my fault hihi)

backend/src/services/auth.rs
```diff
@@ -26,6 +26,7 @@ impl AuthService {
         database: &DatabaseConnection,
         credentials: RegisterModel,
     ) -> Result<HttpResponse, Error> {
+        println!("a");
```
Norbiros, line 29: Remove it

backend/src/main.rs
```diff
@@ -98,7 +98,12 @@ fn setup_actix_app() -> App<
         .service(
             scope("/user")
                 .wrap(AuthMiddleware::with_user())
-                .configure(routes::user::config),
+                .configure(routes::user::user_config),
+        )
+        .service(
```
Norbiros, line 103: Make it a route in `/user`

Norbiros (changes requested):
It looks mostly good

backend/src/main.rs
Kubaryt: i cant do it because it causes errors lol, check dms

backend/src/services/auth.rs
Kubaryt: yass

Norbiros (approved):

goteusz-maszyk (approved):
Ok

PR #187, opened 2025-02-12
Feat: Openapi check in workflows

Kubaryt:
Resolves: #144

PR #188, opened 2025-02-12
feat(backend): Implement db integration tests using sqlite

### LINKED ISSUE #32: [Feat]: Implement integration tests using sqlite
### My suggestion

Implement integration tests for backend (mainly for authorization).

Norbiros:
Depends on #187 

Resolves #32 

More tests, proper mocking etc. will be implemented in separate PR

### DIFF
--- a/backend/src/models/user.rs
+++ b/backend/src/models/user.rs
@@ -5,6 +5,7 @@ use crate::routes::auth::RegisterModel;
 use crate::utils::error::Error;
 use actix_web::dev::Payload;
 use actix_web::{FromRequest, HttpMessage, HttpRequest};
+use chrono::Local;
 use sea_orm::prelude::Uuid as SeaOrmUuid;
 use sea_orm::ActiveValue::Set;
 use sea_orm::QueryFilter;
@@ -106,6 +107,9 @@ impl users::Model {
             username: Set(credentials.name.clone()),
             email: Set(credentials.email.clone()),
             password: Set(Some(password_hash)),
+            created_at: Set(Local::now().naive_local()),
+            is_leader: Set(false),
+            roles: Set(UserRoles::Default),
             ..Default::default()
         }
         .insert(database)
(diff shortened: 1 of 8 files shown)


Kubaryt (changes requested):
properly rebase your branch

backend/src/models/user.rs
```diff
@@ -106,6 +107,9 @@ impl users::Model {
             username: Set(credentials.name.clone()),
             email: Set(credentials.email.clone()),
             password: Set(Some(password_hash)),
+            created_at: Set(Local::now().naive_local()),
+            is_leader: Set(false),
+            roles: Set(UserRoles::Default),
             ..Default::default()
```
Kubaryt, line 113: um why are you defining default values by hand? also if you are already doing that, then why leave ..Default::default() if you cover all fields?

backend/tests/auth.rs
```diff
@@ -85,32 +51,26 @@ async fn register_invalid_email() {
...
-            .app_data(Data::new(AppState::with_database(database)))
-            .service(scope("/auth").configure(routes::auth::config))
-            .service(scope("/user").configure(routes::user::config)),
-    )
-    .await;
+    let app = test::init_service(setup_test_app().await).await;
 
     let register_payload = json!({
         "email": "test@example.com",
-        "password": "okok"
```
Kubaryt, line 113: why :sob::sob::sob::sob:

Kubaryt (changes requested):
thats all… for now

backend/src/models/user.rs
Norbiros: I need to define it here, because default value isn't properly generated by seaorm cli.

Norbiros: No, default is still used here for other fields like `team`

Kubaryt (dismissed):

Kubaryt (approved):

Issue #192, opened 2025-02-14
[Enhancement]: Make CORS less restrictive


Norbiros:
### My idea

It is impossible to allow for all actions from PR deployments (see https://github.com/Hack4Krak/Hack4KrakSite/issues/173), but we may allow for requests to entire /tasks route for everyone

PR #195, opened 2025-02-14
Enhancement: Enabled permissive cors for /tasks endpoints

Kubaryt:
Resolves: #192

### DIFF
--- a/backend/src/lib.rs
+++ b/backend/src/lib.rs
@@ -35,7 +35,7 @@ pub fn setup_actix_app() -> UtoipaApp<
 
     let cors_middleware = Cors::default()
         .allowed_origin_fn(|origin, request| {
-            if request.uri.path().contains("/tasks") {
+            if request.uri.path().starts_with("/tasks") {
                 return true;
             }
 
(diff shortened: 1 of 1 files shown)


backend/src/lib.rs
```diff
@@ -53,7 +63,11 @@ pub fn setup_actix_app() -> UtoipaApp<
                 .wrap(Governor::new(&governor_middleware))
                 .configure(routes::auth::config),
         )
-        .service(scope("/teams").configure(routes::teams::config))
+        .service(
```
Norbiros, line 66: Remove it

Norbiros (changes requested):

backend/src/lib.rs
```diff
@@ -34,8 +34,18 @@ pub fn setup_actix_app() -> UtoipaApp<
         .unwrap();
 
     let cors_middleware = Cors::default()
-        .allowed_origin("http://localhost:3000")
-        .allowed_origin("https://hack4krak.pl")
+        .allowed_origin_fn(|origin, request| {
+            if request.uri.path().contains("/tasks") {
```
Norbiros, line 38: Starts with

Norbiros (changes requested):

backend/src/lib.rs
Kubaryt: valid

Norbiros (approved):

wikipop (approved):

Issue #198, opened 2025-02-14
[Bug]: With team as member method in auth middleware returns 403 if user is not a team leader


Kubaryt:
### Description

If you are not a team leader, but part of team and access for example /my-team you get 403 forbidden you are not team leader

### Steps to reproduce

self explanatory

rena-ertam028:
Hello, we tried to solve the issue.

This is what we did:

Fixed an issue where team members were getting 403 forbidden when accessing /my_team endpoint. The middleware was incorrectly checking for team leader status even when only team member access was required. Modified the auth middleware to only check leader status when TeamRequirement::Leader is specified.

You can review changes in this commit: https://github.com/rena-ertam028/Hack4Krak-Hack4KrakSite-198/commit/86520699df0011785bead680e779dc38cceb69ac.


> [!CAUTION]
> Disclaimer: The concept of solution was created by AI and you should never copy paste this code before you check the correctness of generated code. Solution might not be complete, you should use this code as an inspiration only.

---

Latta AI seeks to solve problems in open source projects as part of its mission to support developers around the world. Learn more about our mission at https://latta.ai/ourmission . If you no longer want Latta AI to attempt solving issues on your repository, you can block this account.

Kubaryt:
Thanks @rena-ertam028 but this issue is already being resolved in #199

PR #199, opened 2025-02-14
Fix: Made auth middleware team requirements return 403 user is not a leader only when team requirement is set to leader

Kubaryt:
Resolves: #198


wikipop (approved):

Norbiros (commented):
If you fix bug like this it would be nice to add tests, to prevent mistakes like this in the future.

Kubaryt:
i can add tests but maybe its better to merge fix now and then write those tests

wojpo (approved):
Write tests later

Issue #182, opened 2025-02-11
[Feat]: Implement tasks refresh endpoint


Norbiros:
### My suggestion

Create an endpoint, only for admins & owners, that refreshes `TasksManager` with new tasks from the fs

Issue #134, opened 2025-02-03
[Feat]: Add managing teams to admin panel


Kubaryt:
### My suggestion

Functionalities:
- view all teams with info
- change team name 
- set team inactive
- remove user from team
- change leader

Issue #133, opened 2025-02-03
[Feat]: Add managing user to admin panel


Kubaryt:
### My suggestion

Managing users:
- changing roles
- changing nicknames
- setting user inactive (also add active boolean field to user model)

PR #201, opened 2025-02-15
Feat: Created admin panel

Kubaryt:
Resolves: #133
Resolves: #134
Resolves: #182 

Depends on #218 

TODO:

- [x] add checking constraints in backend,
- [x] add tasks/refresh
- [x] make sure there is no place in backend in which you can avoid duplication,
- [x] make sure everything adhere to SRP,

> [!IMPORTANT] 
> This is only backend side, also not full, because sending email will have separate pr, same with frontend


backend/src/models/user.rs
```diff
@@ -141,6 +157,72 @@ impl users::Model {
...
+
+        if let Some(username) = update_user_json.username {
+            active_user.username = Set(username);
+        }
+
+        if let Some(email) = update_user_json.email {
+            active_user.email = Set(email);
+        }
+
+        if let Some(team) = update_user_json.team {
```
Kubaryt, line 186: im not sure if changing user team should be part of update user or if it should be add and remove user from team on teams

backend/src/entities/email_confirmation.rs
```diff
@@ -1,8 +1,9 @@
 //! `SeaORM` Entity, @generated by sea-orm-codegen 1.1.4
 
 use sea_orm::entity::prelude::*;
+use serde::{Deserialize, Serialize};
 
-#[derive(Clone, Debug, PartialEq, DeriveEntityModel, Eq)]
+#[derive(Clone, Debug, PartialEq, DeriveEntityModel, Eq, Serialize, Deserialize)]
```
Norbiros, line 6: You added it manually? If I run generator again will it be overridden?

backend/src/models/teams.rs
```diff
@@ -134,6 +147,99 @@ impl teams::Model {
...
+
+            let leader = Self::get_leader(database, id).await?;
+
+            Self::change_leader(database, new_leader, leader).await?;
+        }
+
+        Ok(())
+    }
+
+    pub async fn delete_team(database: &DatabaseConnection, id: Uuid) -> Result<(), Error> {
```
Norbiros, line 232: There is no need to include `_team` in those names

backend/src/models/teams.rs
```diff
@@ -134,6 +147,99 @@ impl teams::Model {
...
+        for member in members {
+            if member.is_leader {
+                return Ok(member);
+            }
+        }
+
+        Err(Error::Team(TeamLeaderNotFound))
+    }
+
+    pub async fn get_teams(database: &DatabaseConnection) -> Result<Vec<TeamWithMembers>, Error> {
```
Norbiros, line 171: Just `list` or something

backend/src/models/teams.rs
```diff
@@ -134,6 +147,99 @@ impl teams::Model {
 
         Ok(())
     }
+
+    pub async fn get_leader(
```
Norbiros, line 151: > Yes, [RFC #344](https://github.com/rust-lang/rfcs/blob/master/text/0344-conventions-galore.md#gettersetter-apis) says that the conventions should be foo for the getter and set_foo for the setter.

Just make it `leader`

Norbiros (changes requested):
I will review more later, I think we can leave this PR as backend part for admin panel and then you can create new one for frontend + integration

backend/src/routes/admin/users/user_list.rs
```diff
@@ -0,0 +1,20 @@
...
+use sea_orm::EntityTrait;
+
+#[utoipa::path(
+    responses(
+        (status = 200, description = "User list successfully retrieved."),
+        (status = 500, description = "Internal server error.")
+    ),
+    tag = "admin/users"
+)]
+#[get("/")]
```
Norbiros, line 15: `/list`

backend/src/routes/admin/users/user_list.rs
```diff
@@ -0,0 +1,20 @@
...
+pub async fn user_list(app_state: Data<AppState>) -> Result<HttpResponse, Error> {
```
Norbiros, line 16: `list(`

backend/src/routes/admin/users/update_user.rs
```diff
@@ -0,0 +1,42 @@
...
+    request_body = UpdateUserModel,
+    responses(
+        (status = 200, description = "User successfully updated."),
+        (status = 403, description = "User must have higher role than updated user."),
+        (status = 500, description = "Internal server error.")
+    ),
+    tag = "admin/users"
+)]
+#[patch("/update/{id}")]
+pub async fn update_user(
```
Norbiros, line 27: `update`

backend/src/models/user.rs
```diff
@@ -141,6 +157,72 @@ impl users::Model {
...
+        active_user.save(database).await?;
+        Ok(())
+    }
+
+    pub async fn delete(
+        database: &DatabaseConnection,
+        user: users::Model,
+        id: SeaOrmUuid,
+    ) -> Result<(), Error> {
+        let deleted_user = users::Entity::find_by_id(id)
```
Norbiros, line 213: `user_to_delete`

backend/src/models/user.rs
```diff
@@ -75,6 +81,16 @@ impl users::Model {
             .one(database)
             .await?;
 
+        if let Some(id) = id {
```
Norbiros, line 84: `if let (Some(id), Some(user)) = (id, user) {`

backend/src/models/teams.rs
```diff
@@ -134,6 +147,99 @@ impl teams::Model {
...
+        Ok(())
+    }
+
+    pub async fn delete(database: &DatabaseConnection, id: Uuid) -> Result<(), Error> {
+        let team = teams::Entity::find_by_id(id)
+            .one(database)
+            .await?
+            .ok_or(Error::Team(TeamNotFound))?;
+
+        let active_team: ActiveModel = team.into();
```
Norbiros, line 238: I'm not sure, but do you have to convert it to active model?

backend/src/models/teams.rs
```diff
@@ -134,6 +147,99 @@ impl teams::Model {
...
+            {
+                return Err(Error::Team(AlreadyExists));
+            }
+            let mut active_team: ActiveModel = team.into();
+            active_team.name = Set(team_name);
+            active_team.update(database).await?;
+        }
+
+        if let Some(leader) = update_team_json.leader {
+            let new_leader = users::Entity::find_by_id(
```
Norbiros, line 217: Add new method `find_by_uuid` for users::Enity

backend/src/models/teams.rs
```diff
@@ -134,6 +147,99 @@ impl teams::Model {
...
+    pub async fn leader(
+        database: &DatabaseConnection,
+        team_id: Uuid,
+    ) -> Result<users::Model, Error> {
+        let team_with_members = teams::Entity::find_by_id(team_id)
+            .find_with_related(users::Entity)
+            .all(database)
+            .await?;
+
+        let members = team_with_members[0].1.clone();
```
Norbiros, line 160: Why .1

backend/src/models/teams.rs
```diff
@@ -134,6 +147,99 @@ impl teams::Model {
...
+
+        for member in members {
```
Norbiros, line 162: .into_iter().find

Norbiros (changes requested):
- Make endpoints match functions
- There is no need to add `_user` or `_team` suffix


Also what about tests :ccc It would be perfect if all endpoints had tests, but just add them for more complicated ones like update entity

backend/src/models/teams.rs
Kubaryt: because its tuple with team and users, but i will change that, cuz now that i think about it it can be done other way

backend/src/models/teams.rs
Kubaryt: by uuid from string

backend/src/models/teams.rs
Kubaryt: i will check that

backend/src/models/teams.rs
```diff
@@ -133,6 +146,92 @@ impl teams::Model {
...
+            {
+                return Err(Error::Team(AlreadyExists));
+            }
+            let mut active_team: ActiveModel = team.into();
+            active_team.name = Set(team_name);
+            active_team.update(database).await?;
+        }
+
+        if let Some(leader) = update_team_json.leader {
+            let new_leader = users::Model::find_by_uuid_from_string(database, &leader).await?;
```
Norbiros, line 213: I don't get this endpoint `find_by_uuid_from_string`, why you can't convert to uuid here? Or even it would be best if route was responsible for doing that. I think serde supports uuid (de)serializaiton

backend/src/models/teams.rs
```diff
@@ -133,6 +146,92 @@ impl teams::Model {
...
+
+    pub async fn delete(database: &DatabaseConnection, id: Uuid) -> Result<(), Error> {
+        let team = teams::Entity::find_by_id(id)
+            .one(database)
+            .await?
+            .ok_or(Error::Team(TeamNotFound))?;
+
+        // Team is declared and then deleted instead of using delete_by_id,
+        // because it should return error if team is not found and delete_by_id does not return this error
+        team.delete(database).await?;
```
Norbiros, line 231: There is this `do_nothing` that returns status reponse

backend/src/models/user.rs
```diff
@@ -61,10 +67,23 @@ impl users::Model {
             .await?)
     }
 
+    pub async fn find_by_uuid_from_string(
```
Norbiros, line 70: Same as above, change name or even remove it

backend/src/models/user.rs
```diff
@@ -61,10 +67,23 @@ impl users::Model {
...
+        database: &DatabaseConnection,
+        id: &str,
+    ) -> Result<Self, Error> {
+        let user = users::Entity::find_by_id(Uuid::parse_str(id).map_err(|_| Error::UserNotFound)?)
```
Norbiros, line 74: Use proper error

Norbiros (changes requested):

backend/src/models/teams.rs
Kubaryt: I was converting here, but you asked me to create separate function for it, also no serde doesnt serialize uuid, because uuid doesnt have serialize and deserialize traits

backend/src/models/user.rs
```diff
@@ -141,6 +155,77 @@ impl users::Model {
...
+        event_config: &EventConfig,
+        id: SeaOrmUuid,
+        update_user_json: UpdateUserModel,
+    ) -> Result<(), Error> {
+        let updated_user = users::Entity::find_by_id(id)
+            .one(database)
+            .await?
+            .ok_or(Error::UserNotFound)?;
+
+        if user.roles.permission_level() <= updated_user.roles.permission_level() {
```
Norbiros, line 171: There is no need to use .permission_level

Norbiros (changes requested):
i think only that

Norbiros (approved):
I didn't review it properly, but it looks mostly good

PR #216, opened 2025-02-18
fix(backend): Include email assets in binary

### LINKED ISSUE #214: [Bug]: Email assets not found on production
### Description

In `Dockerfile` we just binary, and then just copy it to other stage where it is executed. Because of that, no email templates are found. You can update our Dockerfile or just use `include_str!` instead.

### Steps to reproduce

1. Open hack4krak.pl
2. Try to register

_No response_

Norbiros:
Resolves #214


Kubaryt (approved):
it passed tests so i guess it works

Issue #217, opened 2025-02-18
[Bug]: Inviting user to team or accepting invite causes 500 Internal Server Error - Database operation failed


Kubaryt:
### Bug description

So basically when you invite user to team/accept invitation you get this

```sh
2025-02-18T16:28:08.064593Z ERROR hack4krak_backend::middlewares::status_code_drain_middleware: Detected status: 500 Internal Server Error - /teams/management/invite_user - {"Database operation failed"}    
2025-02-18T16:28:08.064623Z DEBUG actix_web::middleware::logger: Error in response: DatabaseOperation(Query(SqlxError(Database(PgDatabaseError { severity: Error, code: "42883", message: "operator does not exist: uuid = text", detail: None, hint: Some("No operator matches the given name and argument types. You might need to add explicit type casts."), position: Some(Original(237)), where: None, schema: None, table: None, column: None, data_type: None, constraint: None, file: Some("parse_oper.c"), line: Some(647), routine: Some("op_error") }))))    
2025-02-18T16:28:08.064724Z  INFO actix_web::middleware::logger: 127.0.0.1 "POST /teams/management/invite_user HTTP/1.1" 500 536 "http://localhost:8080/docs" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/133.0.0.0 Safari/537.36" 0.051175    
```

### Cause

This is caused by part of code in assert_correct_team_size function which is called by those 2 endpoints at some point of their logic.

```rs
let members_count = users::Entity::find()
    .filter(users::Column::Team.eq(name))
    .count(database)
    .await?;
```

This operation compares name, which is name: &str, parameter from this function, to users::Column::Team, which has uuid type, this causes database operation error:

```sh
hack4krak_postgres  | 2025-02-18 16:28:08.064 UTC [100] ERROR:  operator does not exist: uuid = text at character 237
hack4krak_postgres  | 2025-02-18 16:28:08.064 UTC [100] HINT:  No operator matches the given name and argument types. You might need to add explicit type casts.
hack4krak_postgres  | 2025-02-18 16:28:08.064 UTC [100] STATEMENT:  SELECT COUNT(*) AS num_items FROM (SELECT "users"."username", "users"."email", "users"."created_at", "users"."password", "users"."id", "users"."is_leader", "users"."team", CAST("users"."roles" AS text) FROM "users" WHERE "users"."team" = $1) AS "sub_query"
```

[...]

PR #218, opened 2025-02-18
fix: Fixed bug where asserting team size lead to 500 Internal Server Error and added tests to it

Sanielia:
Resolves: #217

### DIFF
--- a/backend/tests/teams.rs
+++ b/backend/tests/teams.rs
@@ -1,17 +1,18 @@
-use utils::{setup_schema, setup_test_app};
+use utils::setup_test_app;
 
+use crate::utils::setup_database_with_schema;
 use actix_web::http::header;
 use actix_web::web::Data;
 use actix_web::{test, App};
 use chrono::{Duration, Local};
 use hack4krak_backend::entities::sea_orm_active_enums::UserRoles;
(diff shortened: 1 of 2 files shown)


backend/tests/teams.rs
```diff
@@ -181,3 +188,132 @@ async fn create_team_success() {
...
     assert_eq!(response.status(), 200);
 }
+
+async fn assert_team_size_setup() -> (DatabaseConnection, Uuid, Uuid) {
+    let database = Database::connect("sqlite::memory:").await.unwrap();
+
+    setup_schema(&database, team_invites::Entity).await;
+    setup_schema(&database, teams::Entity).await;
+    setup_schema(&database, users::Entity).await;
+    setup_schema(&database, email_confirmation::Entity).await;
```
Norbiros, line 198: DRY. Now we have to register all entities for tests 2 times, which  is pretty anoyying

backend/tests/teams.rs
```diff
@@ -181,3 +188,132 @@ async fn create_team_success() {
...
+
+    let users = vec![
```
Norbiros, line 200: Ordering, move this defintion of users just when you need it (so below team creation)

Norbiros (changes requested):

Norbiros (approved):

Issue #176, opened 2025-02-11
[Feat]: Display task's `assets`


Norbiros:
### My suggestion

- [ ] Create an endpoint to retrieve task data (when displaying it on page), as for example full name and assets
- [ ] Properly return paths (names) and descriptions of tasks
- [ ] Create an endpoint to retrieve specific asset

PR #222, opened 2025-02-19
feat: Tasks assets endpoints

Kubaryt:
Resolves: #176


backend/src/routes/task/assets/get.rs
```diff
@@ -0,0 +1,38 @@
...
+    tag = "task/assets"
+)]
+#[get("/get/{task_id}/{asset_path}")]
+pub async fn get(
+    app_state: web::Data<app_state::AppState>,
+    task_asset: web::Path<(String, String)>,
+) -> Result<HttpResponse, Error> {
+    let manager = &app_state.task_manager;
+
+    let task_asset = task_asset.into_inner();
```
Norbiros, line 26: What is .0, and .1 unpack it:
```rs
    let (task_name, asset_id) = task_asset;
```

backend/src/routes/task/assets/get.rs
```diff
@@ -0,0 +1,38 @@
...
+
+    let mut asset_path_full = "assets/".to_string();
+    asset_path_full.push_str(&task_asset.1);
+
+    let content = manager.load_asset(&task_asset.0, &asset_path_full).await?;
+
+    let body = once(ok::<_, Error>(Bytes::from(content)));
```
Norbiros, line 33: There is no real benefit of using streaming then, you are loading entire asset to memory, you could just return it. Please implement streaming from cursor (or something else), that doesn't load everything

Norbiros (changes requested):

backend/src/routes/task/assets/get.rs
```diff
@@ -0,0 +1,36 @@
...
+    tag = "task/assets"
+)]
+#[get("/get/{task_id}/{asset_path}")]
+pub async fn get(
+    request: HttpRequest,
+    task_asset: web::Path<(String, String)>,
+) -> Result<HttpResponse, Error> {
+    let task_asset = task_asset.into_inner();
+
+    let (task_id, asset_path) = (task_asset.0, task_asset.1);
```
Norbiros, line 24: Noo, just `    let (task_id, asset_path) =  task_asset;`

backend/src/routes/task/assets/get.rs
```diff
@@ -0,0 +1,36 @@
...
+
+    let asset_path_full = EnvConfig::get()
+        .tasks_base_path
+        .clone()
+        .join(task_id)
+        .join("assets/")
+        .join(asset_path);
```
Norbiros, line 31: Are you sure all options of pah traversal are prevented? I think it would be best to verify it in manager, and then retrieve. Also if you don't want to modify as much, just change load asset method

Norbiros: Other enpoints can also benefit from this change

Norbiros (changes requested):

backend/src/routes/task/assets/get.rs
Kubaryt: XDDDDDDDDDDDD

backend/src/routes/task/assets/get.rs
```diff
@@ -0,0 +1,31 @@
...
+pub async fn get(
+    app_data: web::Data<AppState>,
+    request: HttpRequest,
+    task_asset: web::Path<(String, String)>,
+) -> Result<HttpResponse, Error> {
+    let manager = &app_data.task_manager;
+
+    let (task_id, asset_path) = task_asset.into_inner();
+
+    let asset_path = "assets/".to_owned() + &*asset_path;
```
Norbiros, line 26: Are you sure you need `&*`? Probably only `&`

Norbiros (changes requested):

backend/src/routes/task/assets/get.rs
Kubaryt: i need &str not &String

Norbiros: .as_str()?

Norbiros (dismissed):

PR #224, opened 2025-02-19
fix(backend): Correctly reset cookies in relaxed security mode

Norbiros:


Kubaryt (approved):
split it to two separate commit

PR #225, opened 2025-02-19
feat(frontend): Team page

Norbiros:


backend/src/routes/teams/management/index.rs
```diff
@@ -0,0 +1,17 @@
...
+        (status = 500, description = "Internal server error."),
+    ),
+    security(
+        ("access_token" = ["leader"])
+    ),
+    tag = "teams/management"
+)]
+#[get("/")]
+pub async fn index() -> Result<HttpResponse, Error>{
+    Ok(HttpResponse::Ok().body("User is team leader"))
```
Norbiros, line 16: Use SuccessResponse

frontend/app/pages/panel/team.vue
```diff
@@ -0,0 +1,94 @@
...
+              {{ user.name }}
+            </div>
+            <UButton v-if="is_user_leader" @click="kick(user.name)">
+              Wyrzuć
+            </UButton>
+          </div>
+          <div v-else class="text-gray-500 flex justify-between items-center">
+            miejsce na uczestnika
+            <UButton v-if="is_user_leader" @click="inviteUserModal = true">
+              Dodaj do drużyny
```
Norbiros, line 87: Make it match designs

frontend/app/pages/panel/team.vue
```diff
@@ -0,0 +1,94 @@
...
+    <div class="flex mx-20 gap-20 mt-10">
+      <div class="border-2 border-neutral-600 p-5 min-w-70 rounded-2xl flex flex-col">
+        <h1 class="flex-grow text-3xl font-bold">
+          {{ team?.team_name }}
+        </h1>
+        <UButton v-if="is_user_leader" class="w-full" @click="deleteTeamModal = true">
+          Usuń zespół
+        </UButton>
+        <UButton v-else class="w-full" @click="leaveTeamModal = true">
+          Opuść zespół
```
Norbiros, line 67: Center text here

frontend/app/pages/panel/team.vue
```diff
@@ -0,0 +1,94 @@
...
+    method: 'DELETE',
+    body: {
+      username,
+    },
+  })
+
+  const toast = useToast()
+
+  if (error.value?.data) {
+    const response = error.value.data as any
```
Norbiros, line 39: I think it would be nice if we could create component, for confirming and eaisly use it here and in other places...

frontend/app/pages/panel/team.vue
```diff
@@ -0,0 +1,94 @@
...
+const { data: team, refresh } = await useAuth('/teams/membership/my_team', {
+  key: 'teams-membership-my-team',
+  method: 'GET',
+})
+
+if (!team.value) {
+  await navigateTo('/panel/')
+}
+
+const { _, error } = await useAuth('/teams/management/', {
```
Norbiros, line 11:
```suggestion
const { error } = await useAuth('/teams/management/', {
```

frontend/app/components/panel/modal/InviteUser.vue
```diff
@@ -0,0 +1,56 @@
...
+
+const state = reactive<Partial<Schema>>({
+  name: undefined,
+})
+
+const toast = useToast()
+const open = defineModel<boolean>()
+const formRef = useTemplateRef('form')
+
+// ToDo: Use $auth isntead of useAuth in fucntions
```
Norbiros, line 19: solve it

Norbiros (commented):
```
 ERROR   ✘  Every operation must have a unique operationId. at #/paths/~1user~1/get/index                                                                                10:19:57 AM


 ERROR  Could not compile template types/open-fetch/schemas/auth.ts.                                                                                                nuxt 10:19:57 AM


 ERROR  Every operation must have a unique operationId. at #/paths/~1user~1/get/index     
 
 ```
 
I'll work on it later... Also add option to view sent invitations. Maybe we can just display slot as "taken" but display that invitation is sent but user didn't accept it yet?

frontend/app/components/panel/modal/InviteUser.vue
Kubaryt: I DONT EVEN KNOW WHY, HOW AND WHATS THE DIFFERENCE

Norbiros: The same i did in /login. `useFetch` is wrapper around `$fetch` to use in ssr context, but here if you run code when user clicks, there is no need to do that, it is always and only done only on the client, so just use `$fetch` instead. Here we are using openapi, so instead use `$auth`

Issue #196, opened 2025-02-14
[Feat]: Account deletion


goteusz-maszyk:
### My suggestion

People should be allowed to remove their data.

Kubaryt:
Frontend side will be implemented in #240

Issue #240, opened 2025-02-20
[Feat]: Add account management to panel


Kubaryt:
### My suggestion

Basically add this to panel

![Image](https://github.com/user-attachments/assets/ff1d1da0-c519-422c-8fad-0e7ae8ae5cce)

but without sessions, because for now we dont use them and with update username or email

PR #241, opened 2025-02-20
feat: Added delete account endpoint

Kubaryt:
Resolves: #196 

> [!IMPORTANT]
> This is only backend side, frontend side will be implemented in #240


backend/src/lib.rs
```diff
@@ -65,6 +66,11 @@ pub fn setup_actix_app(
         .service(scope("/tasks").configure(routes::task::config))
         .service(scope("/user").configure(routes::user::config))
         .service(scope("/event").configure(routes::event::config))
+        .service(
```
Norbiros, line 69: I think you can merge that with /user and in future we can rename it to /account

Norbiros (changes requested):

Norbiros (changes requested):

Norbiros (approved):

Issue #215, opened 2025-02-18
[Refactor]: Use proper library for placeholders


Norbiros:
### My idea

There are a lot of rust based libraries for parsing placeholders. There is no need to create wacky and in-secure reimplementation. For example now if username is `%username%` we are going to replace it

PR #251, opened 2025-02-21
enhancement: Changed parsing placeholders to text template library

Kubaryt:
Resolves: #215


backend/Cargo.toml
```diff
@@ -58,6 +58,7 @@ oauth2 = { version = "5.0.0", features = ["reqwest"] }
 lettre = "0.11.13"
 dashmap = "6.1.0"
 validator = { version = "0.20.0", features = ["derive"] }
+text-template = "0.1.0"
```
Norbiros, line 61: Do we really want to want use unmaintained library not updated for 6 years, with no readme hosted on gitlab. I was thinking about something more powerful with templates, so we could easily create like components etc

Norbiros (changes requested):

backend/Cargo.toml
Norbiros: I was thinking about libraries like tera or even handlebars `handlebars` with html escaping, basic logic like if statements, templates etc.

Issue #242, opened 2025-02-20
[Feat]: Add regulations page


Kubaryt:
### My suggestion

if you need regulations in markdown, then ask me on dm or here

PR #255, opened 2025-02-22
feat: Added rules page

Kubaryt:
Resolves: #242


frontend/app/assets/md/Regulamin_Hack4Krak_LO31.md
Norbiros, line 1: We don't have folders like `png`, `svg` so why `md`? Name it `docs`. Also change files to something like `School_Edition_Guidelines.md`

frontend/app/pages/rules.vue
```diff
@@ -0,0 +1,11 @@
+<script setup lang="ts">
+import file from 'assets/md/Regulamin_Hack4Krak_LO31.md?raw'
+
+const regulations = file
+</script>
+
+<template>
+  <div class="flex w-full items-center justify-center mt-5">
```
Norbiros, line 8: I think it can be a bit wider

frontend/app/pages/rules.vue
```diff
@@ -0,0 +1,11 @@
+<script setup lang="ts">
```
Norbiros, line 1: Configure in `nuxt.config.ts` `prerender`: `true`

frontend/app/pages/rules.vue
```diff
@@ -0,0 +1,11 @@
...
+import file from 'assets/md/Regulamin_Hack4Krak_LO31.md?raw'
+
+const regulations = file
```
Norbiros, line 4: maybe just use `as` keyword in import? I'm not sure if it will work

frontend/nuxt.config.ts
```diff
@@ -84,6 +84,14 @@ export default defineNuxtConfig({
     preference: 'dark',
     storageKey: 'nuxt-color-mode-forced',
   },
+  mdc: {
+    remarkPlugins: {
+      'remark-math': {},
+    },
+    rehypePlugins: {
+      'rehype-mathjax': {},
```
Norbiros, line 92: Why both? What is the difference between them?

Norbiros (changes requested):

frontend/app/pages/rules.vue
Kubaryt: i wil try

frontend/nuxt.config.ts
Kubaryt: idk but people were using both of them at same time, and if i didnt have one of those then it didnt work

Norbiros: okay lol

Norbiros (changes requested):
Also i think it would fit better in `content` directory

frontend/app/components/MarkdownContent.server.vue
```diff
@@ -1,11 +1,12 @@
 <script setup lang="ts">
 const props = defineProps<{
   text: string
+  width: string | undefined
```
Norbiros, line 4: Good, please no. Make it work like nuxt ui does that, because what if i want to add for example height? new property?

Norbiros (changes requested):

frontend/app/components/MarkdownContent.server.vue
Kubaryt: wdym?

Norbiros (approved):

PR #261, opened 2025-02-23
enhancement(frontend): Display time left to the event

Norbiros:


Kubaryt:
no approves… ban

Norbiros:
@Mr Czapla said "norbert", so I think that is enough for merge

PR #275, opened 2025-02-24
fix(frontend) Use `$auth` instead of `useAuth` in client events

Norbiros:


Kubaryt (approved):

Kubaryt (changes requested):
Rebase with master so typechecks will run

LupusVirtute (commented):
No idea why `$` because $ usually in javascript signifies its an object and youre using it here as a function

frontend/app/components/panel/modal/CreateTeam.vue
```diff
@@ -16,18 +16,21 @@ const toast = useToast()
...
+  const response = await $auth('/teams/create', {
     method: 'POST',
     body: {
       team_name: event.data.name,
     },
   })
 
-  if (error.value?.data === undefined) {
-    toast.add({ title: 'Sukces', description: 'Pomyślnie stworzono team', color: 'success' })
+  if ((response as any).error) {
```
Kubaryt, line 29: is $auth displaying toast on error? if not display it, also in other cases

Norbiros: yes, it displays

PR #279, opened 2025-02-24
fix(backend): Use proper default env config for tests

### LINKED ISSUE #277: [Bug]: Default doesn't work properly for envs in load_test_config method
### Description

All envs except for jwt secret which is explicitly set in this config is just unit string ""

### Steps to reproduce

1. Open any test which loads test config
2. Print any env with declared default except for jwt secret
3. Make test panic so you see output

_No response_

Norbiros:
Resolves #277

### DIFF
--- a/backend/src/routes/teams/management/mod.rs
+++ b/backend/src/routes/teams/management/mod.rs
@@ -8,7 +8,8 @@ pub mod rename;
 mod revoke_invitation;
 
 pub fn config(config: &mut utoipa_actix_web::service_config::ServiceConfig) {
-    config.service(change_leader::change_leader)
+    config
+        .service(change_leader::change_leader)
         .service(invite_user::invite_user)
         .service(kick_user::kick_user)
         .service(rename::rename)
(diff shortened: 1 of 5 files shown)


goteusz-maszyk (changes requested):
Replace all defaults with ..default::default()

Norbiros:
It doesn't work like that mate

goteusz-maszyk:
ok then fix conflicts

LupusVirtute (approved):
LGTM

backend/src/routes/teams/management/mod.rs
```diff
@@ -8,7 +8,8 @@ pub mod rename;
 mod revoke_invitation;
 
 pub fn config(config: &mut utoipa_actix_web::service_config::ServiceConfig) {
-    config.service(change_leader::change_leader)
+    config
+        .service(change_leader::change_leader)
```
Kubaryt, line 12: unrelated but ok ig

Kubaryt (approved):
LGTM

Issue #263, opened 2025-02-23
[Feat]: Add team confirmation


Kubaryt:
### My suggestion

- [x]  Add `is_confirmed` flag and confirmation code uuid field to team
- [x]  Add endpoint for team confirmation

Norbiros:
I think status would be better (enum).

PR #293, opened 2025-02-25
feat: Added team confirmation

Kubaryt:
Resolves: #263


backend/src/entities/sea_orm_active_enums.rs
```diff
@@ -1,15 +1,26 @@
-//! `SeaORM` Entity, @generated by sea-orm-codegen 1.1.5
+//! `SeaORM` Entity, @generated by sea-orm-codegen 1.1.6
 
 use sea_orm::entity::prelude::*;
 use serde::{Deserialize, Serialize};
+use utoipa::ToSchema;
```
Norbiros, line 5: Ummm, it will be overidden every time? Use `impl`

backend/src/models/teams.rs
```diff
@@ -264,6 +269,23 @@ impl teams::Model {
...
+        database: &DatabaseConnection,
+        confirmation_code: Uuid,
+    ) -> Result<(), Error> {
+        let team = teams::Entity::find()
+            .filter(teams::Column::ConfirmationCode.eq(confirmation_code))
+            .one(database)
+            .await?
+            .ok_or(Error::Team(TeamNotFound))?;
+
+        let mut active_team: ActiveModel = team.into();
```
Norbiros, line 283: Model should be responsible for logic related to itself. Create a service or run 2 functions in route

backend/src/routes/teams/mod.rs
```diff
@@ -76,7 +80,8 @@ impl error::ResponseError for TeamError {
             TeamError::TeamNotFound
             | TeamError::UserDoesntHaveAnyInvitations
             | TeamError::UserDoesntHaveInvitationsFromTeam { .. }
-            | TeamError::TeamLeaderNotFound => StatusCode::NOT_FOUND,
+            | TeamError::TeamLeaderNotFound
+            | TeamError::InvalidConfirmationCode => StatusCode::NOT_FOUND,
```
Norbiros, line 84: I don't think not found is proer error code

Norbiros (changes requested):
And how we can generate confirmation codes for team, and easily view them to create qr codes?

backend/src/entities/sea_orm_active_enums.rs
Kubaryt: how

Norbiros: look how macro does that, or pass a parameter to seaorm cli

Kubaryt: there is no parameter related to utoipa to schema in seaormcli

Norbiros (changes requested):
Tests 👉👈

Issue #307, opened 2025-02-27
[Enhancement]: Allow only confirmed teams to submit flags


Kubaryt:
### My idea

as follows

PR #309, opened 2025-02-27
enhancement: Disalllowed unconfirmed teams from submiting flag

Kubaryt:
Resolves: #307

### DIFF
--- a/backend/src/routes/flag/mod.rs
+++ b/backend/src/routes/flag/mod.rs
@@ -17,7 +17,7 @@ pub enum FlagError {
     InvalidFlag,
     #[error("This team already submitted this flag")]
     AlreadySubmittedFlag,
-    #[error("This team is not confirmed yet")]
+    #[error("This endpoint is only accessible for confirmed teams")]
     TeamNotConfirmed,
 }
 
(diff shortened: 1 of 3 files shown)


backend/src/routes/flag/submit.rs
```diff
@@ -34,6 +35,10 @@ pub async fn submit(
     model: Validated<Json<SubmitModel>>,
     team: teams::Model,
 ) -> Result<HttpResponse, Error> {
+    if team.status != TeamStatus::Confirmed {
```
Norbiros, line 38: In future we probably should probably create middleware for that :3

backend/src/routes/flag/mod.rs
```diff
@@ -17,13 +17,16 @@ pub enum FlagError {
     InvalidFlag,
     #[error("This team already submitted this flag")]
     AlreadySubmittedFlag,
+    #[error("This team is not confirmed yet")]
```
Norbiros, line 20: Maybe something like `This endpoint is only accessible for confirmed teams`

Norbiros (changes requested):

Norbiros (approved):

PR #371, opened 2025-03-12
enhancement: Remove kebab case from event config

### LINKED ISSUE #311: [Enhancement]: Remove kebab case from serialization in task models
### My idea

its currently used because of how fields in eventconfig are defined in yaml

Norbiros:
Resolves #311


LupusVirtute:
Not an enhancement but refactor but LGTM

Issue #364, opened 2025-03-11
[Enhancement]: Use polish in email sender's name


Norbiros:
### My idea

Mails are written in polish, but sender's name look like `Hack4Krak Authentication`. https://github.com/Hack4Krak/Hack4KrakSite/pull/276#discussion_r1990049984

PR #401, opened 2025-03-26
enhancement: Use polish names for sending emails

Kubaryt:
Resolves: #364

### DIFF
--- a/backend/src/services/auth.rs
+++ b/backend/src/services/auth.rs
@@ -46,7 +46,7 @@ impl AuthService {
 
         Email {
             sender: (
-                Some("Hack4Krak Authentication".to_string()),
+                Some("Autoryzacja Hack4Krak".to_string()),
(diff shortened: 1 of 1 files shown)


backend/src/services/auth.rs
```diff
@@ -46,8 +46,8 @@ impl AuthService {
 
         Email {
             sender: (
-                Some("Hack4Krak Authentication".to_string()),
-                "auth@hack4krak.pl".to_string(),
+                Some("Autoryzacja Hack4Krak".to_string()),
+                "autoryzacja@hack4krak.pl".to_string(),
```
Norbiros, line 50: I think we can leave email as it is. I couldn't find any site that uses non-english in email address

Norbiros (changes requested):
Please, add `backend` scope to the commit name

Norbiros (approved):

Issue #365, opened 2025-03-11
[Enhancement]: Remove hardcoded domain


Norbiros:
### My idea

In a lot of places we have hardcoded `hack4krak.pl`. Extract it to env config and use it properly

PR #402, opened 2025-03-26
enhancement: Remove hardcoded domain

Kubaryt:
Resolves: #365


backend/src/services/env.rs
```diff
@@ -35,11 +37,12 @@ impl Default for EnvConfig {
         EnvConfig {
             relaxed_security_mode: false,
             backend_address: "127.0.0.1:8080".to_string(),
-            oauth_finish_redirect_url: "http://localhost:3000/panel".to_string(),
+            backend_domain: "http://localhost:8080".to_string(),
```
Norbiros, line 40: > A domain name consists of one or more parts, technically called labels, that are conventionally concatenated, and delimited by dots, such as [example.com](https://en.wikipedia.org/wiki/Example.com).

This ain't domain

Norbiros (changes requested):
BTW maybe we can use URL crate, for easier and more secure joining

backend/src/services/env.rs
Norbiros: Nope. Rename to `url`, `address` or something like that

backend/src/services/auth.rs
```diff
@@ -42,13 +42,13 @@ impl AuthService {
         let confirmation_code =
             email_confirmation::Model::create_with_userinfo(&app_state.database, user_info).await?;
 
-        let confirmation_link = Self::create_email_confirmation_link(&confirmation_code);
+        let confirmation_link = Self::create_email_confirmation_link(&confirmation_code)?;
+
+        let mut sender_email = "auth@".to_string();
+        sender_email.push_str(&EnvConfig::get().domain_name);
```
Norbiros, line 48: Why not format?

backend/src/services/env.rs
```diff
@@ -14,19 +15,17 @@ pub struct EnvConfig {
     pub relaxed_security_mode: bool,
     pub database_url: String,
     pub backend_address: String,
-    pub oauth_finish_redirect_url: String,
+    pub domain_name: String,
```
Norbiros, line 18: Just domain

backend/src/utils/openapi.rs
```diff
@@ -16,13 +16,20 @@ use utoipa::openapi::security::{ApiKey, ApiKeyValue, SecurityScheme};
...
     modifiers(&SecurityAddon),
-    servers(
-        (url = "http://localhost:8080", description = "Local development server"),
-        (url = "https://api.hack4krak.pl", description = "Main API server"),
-    )
 )]
 pub struct ApiDoc;
 
+impl ApiDoc {
+    pub fn openapi_with_server() -> OpenApi {
```
Norbiros, line 23: `with_server`

Norbiros (changes requested):
Also update backend's readme.md

Norbiros (approved):
LGTM, but I'll merge it tomorrow, because I have to configure env variables

Issue #249, opened 2025-02-21
[Enhancement]: Make some modules private


Norbiros:
### My idea

Now almost all modules in binary are public, which means checks for unsued methods, classes etc. don't run, which is pretty annoying

PR #406, opened 2025-03-27
refactor: Made some modules private

Kubaryt:
Resolves: #249


Norbiros (changes requested):
1. Sorry I named the issue wrongly, it should be `refactor`
2. You could also look into lib.rs file

Kubaryt:
> 1. Sorry I named the issue wrongly, it should be `refactor`
> 2. You could also look into lib.rs file

ad 2.
i looked into it and all of them should be public, because otherwise compilation on precommit fails, tho I will recheck that

Norbiros:
You may have to use `pub(crate)`

Kubaryt:
Other modules are used in main binary or tests, so pub(crate) doesn't work there, should I also use pub(crate) with submodules?

Norbiros (approved):
There isn't any good way to run checks for this unused stuff (to spot for example those errors)... If you have any question about my changes ask here or on Discord.

Please correctly name and squash commits before merging

Issue #248, opened 2025-02-21
[Enhancement]: Create a script for writing openapi specification


Norbiros:
### My idea

Now you have to start the server, which requires database connection etc. Create simple script that works like "is openapi up to date" test, that just instead of comparing writes new file.

PR #407, opened 2025-03-27
feat: Added openapi generation cli

Kubaryt:
Resolves: #248


backend/openapi-cli/src/main.rs
```diff
@@ -0,0 +1,29 @@
...
+async fn verify_openapi(openapi: String) {
+    let generated_openapi = read_to_string("../frontend/openapi/api/openapi.json")
+        .await
+        .unwrap();
+    assert_eq!(openapi, generated_openapi, "Your openapi json is...");
+    println!("OpenAPI specification is correct!");
+}
+
+async fn write_openapi(openapi: String) {
+    tokio::fs::write("../frontend/openapi/api/openapi.json", openapi)
```
Norbiros, line 25: Extract path to const

Norbiros (changes requested):
Mention it in backend teadmd

backend/openapi-cli/src/main.rs
Kubaryt: maybe env?

Norbiros (approved):

PR #494, opened 2025-04-27
feat(frontend): Redesign frontend `Dockerfile`

Norbiros:


frontend/Dockerfile
```diff
@@ -1,40 +1,26 @@
...
 
-# install dependencies into temp directory
-# this will cache them and speed up future builds
-FROM oven/bun:1.2-alpine AS install
-WORKDIR /usr/src/app
+FROM base as install
+COPY package.json bun.lock .
+COPY frontend/package.json frontend/
+COPY backend/package.json backend/
+COPY patches/ patches/
```
wikipop, line 8: why are we layering this?

frontend/Dockerfile
```diff
@@ -1,40 +1,26 @@
...
+RUN bun run build
 
-## copy production dependencies and source code into final image
 FROM base AS release
-COPY --from=prerelease /usr/src/app/node_modules node_modules
-COPY --from=prerelease /usr/src/app/.output .output
-COPY --from=prerelease /usr/src/app/package.json .
+COPY --from=prerelease /app/node_modules node_modules
+COPY --from=prerelease /app/.output .output
+COPY --from=prerelease /app/package.json .
```
wikipop, line 23: and this?

wikipop (approved):

frontend/Dockerfile
Norbiros: Because patches affect installed libraries in `node_modules`

frontend/Dockerfile
Norbiros: It doesn't weight much and it is nice to have an easy way to start it using for example `bun run preview` if you want

PR #505, opened 2025-04-29
enhancement(frontend): Use `katex` for math formulas in Markdown

Norbiros:
It produces inline math formulas in html, so users can easily copy it. And it is faster and way more lightweight :3

It should not produce those annoying warnings...


wikipop (approved):
lgtm; havn't tested

Issue #403, opened 2025-03-26
[Enhancement]: Include response body in all utoipa::path configs


Kubaryt:
### My idea

So in some endpoints response body is not declared, declare it

PR #541, opened 2025-05-06
enhancement: Add response body to all endpoints which have it missing

Kubaryt:
Resolves: #403 

honestly i thought it would be more lol


backend/src/entities/prelude.rs
```diff
@@ -1,7 +1,6 @@
-//! `SeaORM` Entity, @generated by sea-orm-codegen 1.1.6
+//! `SeaORM` Entity, @generated by sea-orm-codegen 1.1.10
 
 pub use super::email_confirmation::Entity as EmailConfirmation;
-pub use super::flag_capture::Entity as FlagCapture;
```
Norbiros, line 4: um

Norbiros (changes requested):

backend/src/entities/prelude.rs
Kubaryt: kms

Norbiros (approved):

PR #544, opened 2025-05-07
feat(frontend): Use primary background color for `ElevatedButton`

Norbiros:
image

Approved by the best designer in our team :3


Kubaryt:
…wdym by whole crabcraft backend

Kubaryt (approved):
this feels bright af, but i prefer it over this random red instead of primary (i was going to ask why there is random red, when everything else on panel is yellow/black/white)

Kubaryt (approved):

PR #548, opened 2025-05-08
feat: Unified email verification

### LINKED ISSUE #547: [Refactor]: Create universal email verfication request system
### What to refactor?

Currently we have 2 different tables:
- password_reset
- email_verification

They have a lot of simillar logic, we could combine them into big one

Norbiros:
Resolves #547


backend/locales/error.yaml
```diff
@@ -31,6 +31,8 @@ Error:
   FailedToParseUrl: "Nie udało się sparsować URL {0}"
   ServerEventSendingError: "Błąd wysyłania zdarzenia serwera {0}"
   Validator: "Niepoprawne pola w zapytaniu"
+  InvalidEmailConfirmationCode: "Nieprawidłowy kod z e-mail"
+  EmailConfirmationCodeExpired: "Kod z tego e-mail wygasł. Spróbuj ponownie!"
```
Kubaryt, line 35: image
remove duplicated errors

backend/package.json
```diff
@@ -6,9 +6,14 @@
     "build": "cargo build",
     "dev": "cargo run",
     "lint": "cargo fmt -- --check && cargo clippy",
-    "lint:fix": "cargo fmt && cargo clippy --fix --allow-staged --allow-dirty",
+    "lint:fix": "cargo clippy --fix --allow-staged --allow-dirty && cargo fmt",
```
Kubaryt, line 9: is there some convention about command i don't know?

backend/DEVELOPMENT.md
```diff
@@ -100,28 +100,28 @@ bun openapi-cli -- write
...
   ```
-
-> [!NOTE]  
-> All the following commands should be run from the `backend/migration/` directory
-> You have to set the `DATABASE_URL` environment variable before.
-
 - Generate a new migration file
   ```sh
-  cargo run -- generate MIGRATION_NAME
+  bun db:migration <MIGRATION_NAME>
```
Kubaryt, line 114: maybe
```suggestion
  bun db:migration <migration_name>
```
cuz we name migrations in snake_case?

backend/src/models/email_verification_request.rs
```diff
@@ -0,0 +1,130 @@
...
+
+        email_verification_request::Entity::insert(email_verification_request::ActiveModel {
+            id: Set(confirmation_code),
+            action_type: Set(action_type),
+            additional_data: Set(action_data),
+            email: Set(email),
+            expiration_time: Set(
+                expiration_duration.map(|duration| Utc::now().naive_utc().add(duration))
+            ),
+            created_at: Set(chrono::Utc::now().naive_utc()),
```
Kubaryt, line 44:
```suggestion
            created_at: Set(Utc::now().naive_utc()),
```
cuz' you are using it without prefix above (also consider importing Utc and Duration :3)

backend/src/models/email_verification_request.rs
```diff
@@ -0,0 +1,130 @@
...
+        })
+        .exec(database)
+        .await?;
+
+        Ok(confirmation_code)
+    }
+
+    pub async fn get(database: &DatabaseConnection, id: Uuid) -> Result<Self, Error> {
```
Kubaryt, line 52:
```suggestion
    pub async fn find_by_id(database: &DatabaseConnection, id: Uuid) -> Result<Self, Error> {
```
or something, because find by id is entity method, anyway it should match names in other struct, at lest this is what you said to me :3

backend/src/models/email_verification_request.rs
```diff
@@ -0,0 +1,130 @@
...
+        let (action_type, action_data) = action.get();
+        let confirmation_code = Uuid::new_v4();
+
+        email_verification_request::Entity::insert(email_verification_request::ActiveModel {
+            id: Set(confirmation_code),
+            action_type: Set(action_type),
+            additional_data: Set(action_data),
+            email: Set(email),
+            expiration_time: Set(
+                expiration_duration.map(|duration| Utc::now().naive_utc().add(duration))
```
Kubaryt, line 42: does it handle expiration duration being None?

backend/src/models/email_verification_request.rs
```diff
@@ -0,0 +1,130 @@
...
+
+                return Err(Error::EmailConfirmationCodeExpired);
+            }
+        }
+
+        Ok(email_verification_request)
+    }
+
+    pub async fn delete(self, database: &DatabaseConnection) -> Result<(), Error> {
+        let active_password_reset: email_verification_request::ActiveModel = self.into();
```
Kubaryt, line 72:
```suggestion
        let active_email_verifiaction_request: email_verification_request::ActiveModel = self.into();
```

backend/src/routes/auth/confirm.rs
```diff
@@ -19,7 +20,8 @@ pub async fn confirm_email(
     app_state: web::Data<app_state::AppState>,
     confirmation_code: web::Path<String>,
 ) -> Result<HttpResponse, Error> {
-    match AuthService::confirm_email(&app_state, confirmation_code.into_inner()).await {
+    let confirmation_code_uuid = Uuid::parse_str(&confirmation_code.into_inner()).unwrap();
```
Kubaryt, line 23: one question mr Norbiros, why ~~so serious~~ use unwrap?

backend/src/services/auth.rs
```diff
@@ -113,21 +115,19 @@ impl AuthService {
 
     pub async fn confirm_email(
         app_state: &app_state::AppState,
-        confirmation_code: String,
+        confirmation_code: Uuid,
```
Kubaryt, line 118: why confirmation code was string btw if it was always uuid xdddd (maybe i implemented that and for some reason it was like that and no one will ever know why)

backend/src/models/email_verification_request.rs
```diff
@@ -0,0 +1,130 @@
...
+use sea_orm::{DatabaseConnection, EntityTrait, Set};
+use serde::{Deserialize, Serialize};
+use serde_json::Value;
+use uuid::Uuid;
+
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(rename_all = "snake_case", tag = "name", content = "data")]
+pub enum EmailVerificationAction {
+    ConfirmEmailAddress { user_information: UserInformation },
+    ResetPassword { email: String },
```
Kubaryt, line 13: why we are keeping email in it if its also kept in EmailVerificationRequest?
you aint even using it in password reset method xdd
image

backend/src/utils/error.rs
```diff
@@ -65,6 +65,8 @@ pub enum Error {
     FailedToSendEmail(#[from] lettre::transport::smtp::Error),
     FailedToBuildEmail(#[from] lettre::error::Error),
     InvalidJson(#[from] serde_json::Error),
+    InvalidEmailConfirmationCode,
+    EmailConfirmationCodeExpired,
```
Kubaryt, line 69: remove duplications of those from routes::auth::mod

backend/tests/routes/auth.rs
```diff
@@ -146,18 +146,22 @@ async fn auth_flow() {
...
+        id: Set(confirmation_code),
         email: Set("".to_string()),
-        user_info: Set(json![{
-            "name": "test_user",
-            "email": "example@gmail.com",
-            "password_hash": "$argon2id$v=19$m=19456,t=2,p=1$nTzWdmrtGEOnwCocrg76xg$yv16FfDT5+meKwPmSiV+MF9kP8Man6bXZs+BloFTKIk".to_string(),
-        }]),
-        code: Set(confirmation_code.clone()),
-        expiration_date: Set(Utc::now().naive_utc() + chrono::Duration::minutes(30)),
+        action_type: Set("confirm_email_address".to_string()),
```
Kubaryt, line 153: you can use reset password here, because it has less data in it (you did the same in unit tests

Kubaryt (changes requested):
run bun generate:entities
because Defualt macro was added only to email_verification_request

backend/src/models/email_verification_request.rs
Norbiros: This method does a lot more than only finding by id, but I agree that `get` isn't the best name

backend/src/models/email_verification_request.rs
Norbiros: Yeah, if it is none we are not verifying it

backend/package.json
Norbiros: Clippy doesn't care about newlines, empty spaces etc. So on some very rare occasions it may break something that formatter will fix

backend/src/models/email_verification_request.rs
Kubaryt: so if you give expierience_duration=None to create function, it will set expiration time to none? doesn't adding None value to utc value break anything?

backend/tests/routes/auth.rs
Norbiros: useless, it won't hurt to test more

backend/src/models/email_verification_request.rs
Norbiros: If expiration_duration is None, the map just returns None, so nothing is added — result is None.

If it's Some(duration), it adds it to Utc::now().naive_utc().

backend/locales/error.yaml
```diff
@@ -31,13 +31,14 @@ Error:
...
 
 AuthError:
   UserAlreadyExists: "Użytkownik już istnieje"
   InvalidCredentials: "Nieprawidłowa nazwa użytkownika i/lub hasło"
   InvalidEmailAddress: "Nieprawidłowy adres e-mail"
   PasswordAuthNotAvailable: "Uwierzytelnianie hasłem i e-mailem jest niedostępne dla tego konta"
-  InvalidConfirmationCode: "Nieprawidłowy kod potwierdzenia"
   ConfirmationCodeExpired: "Kod potwierdzenia wygasł, spróbuj wysłać e-mail ponownie"
```
Kubaryt, line 42: this also should be removed i think

backend/src/routes/auth/mod.rs
```diff
@@ -32,7 +32,6 @@ pub enum AuthError {
     InvalidCredentials,
     InvalidEmailAddress,
     PasswordAuthNotAvailable,
-    InvalidConfirmationCode,
     ConfirmationCodeExpired,
```
Kubaryt, line 35: this shouldalso be removed i think

Kubaryt (changes requested):

Kubaryt (approved):

Issue #515, opened 2025-04-30
[Feat]: Dynamic navbar


Norbiros:
### My suggestion

If user is logged in display `Otwórz panel`, else `Start gry' or something like that

PR #557, opened 2025-05-12
feat(frontend): Made text on button on Navbar dynamic

Kubaryt:
Resolves: #515

### DIFF
--- a/frontend/app/components/Navbar.vue
+++ b/frontend/app/components/Navbar.vue
@@ -9,6 +9,8 @@ const { data: isLoggedIn } = useAuth('/account/', {
   transform: (data) => {
     return data !== undefined
   },
+  key: 'isLoggedIn',
+  redirect: 'error',
   onResponseError: undefined,
 })
 
--- a/frontend/app/components/Navbar.vue
+++ b/frontend/app/components/Navbar.vue
@@ -4,6 +4,14 @@ import { NAVBAR_ITEMS } from '~~/content/navbar'
 const [DefineNavbarTemplate, ReuseNavbarTemplate] = createReusableTemplate()
 
 const isMobileMenuOpen = ref(false)
+
+const { data: isLoggedIn } = useAuth('/account/', {
+  transform: (data) => {
+    return data !== undefined
+  },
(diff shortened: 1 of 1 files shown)


frontend/app/components/Navbar.vue
```diff
@@ -4,6 +4,15 @@ import { NAVBAR_ITEMS } from '~~/content/navbar'
 const [DefineNavbarTemplate, ReuseNavbarTemplate] = createReusableTemplate()
 
 const isMobileMenuOpen = ref(false)
+
+const { data } = useAuth('/account/', {
+  onResponseError: undefined,
+})
+
+const isLoggedIn = computed(() => {
+  return data.value !== undefined
```
Norbiros, line 13: Use transform

Norbiros (changes requested):
Try making navbar a server component for performance reasons

frontend/app/components/Navbar.vue
Kubaryt: wdym? like transform in css or some nuxt thing i don't kniw about or what?

Norbiros: yup, in usefetch

Norbiros (changes requested):
```
[`Server components(and islands)`] "Navbar" must have a single root element. (HTML comments are considered elements as well.)
```

I also get a lot of hydration errors, but i think this is not because of this pr

Kubaryt:
so problem is with navbar being server component, when its one, then mobile navbar doesn't work

Kubaryt:
…and tests still fail

Kubaryt:
im not sure what the fuck is happening but this happens when I try to build it, even tho i didn't change a single thing related to this page, also I checked and if I comment out lines with fetchin /account/ then i don't get this error…
image

Kubaryt:
What is even more confusing, that navbar-mobile.test.ts passes on local dev, but basic.test.ts fails due to reason I cannot comprehend, I double-checked and literally only h1 tag on index is fucking Hack4Krak text…
image

Norbiros (approved):

PR #562, opened 2025-05-13
fix: Add responsivnes to centered layout and description to personal info form

Kubaryt:
Resolves: @Norbiros death threats

### DIFF
--- a/frontend/app/components/AuthForm.vue
+++ b/frontend/app/components/AuthForm.vue
@@ -77,7 +77,7 @@ async function onSubmit(event: FormSubmitEvent<Schema>) {
 </script>
 
 <template>
-  <div class="sm:w-90 w-60">
+  <div>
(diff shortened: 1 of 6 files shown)


frontend/app/components/AuthForm.vue
```diff
@@ -77,7 +77,7 @@ async function onSubmit(event: FormSubmitEvent<Schema>) {
 </script>
 
 <template>
-  <div class="w-90">
+  <div class="sm:w-90 w-60">
```
Norbiros, line 80: So you are sure it works on all widths, why we can't define something like max width and add padding so it always scales properly

Kubaryt: i don't really know what are you talking about so i will say taht i am sure

Norbiros (approved):

PR #577, opened 2025-05-21
fix: Fetch new auth status endpoint on navbar to decrease number of data

Kubaryt:


Norbiros (changes requested):
Now I always get `Otwórz panel`, please write tests or something, because I don't want to see 3rd PR for this issue.

Additionally I think by default (if backend is not accessible etc) we should display "Zaloguj się"

Kubaryt:
> Now I always get `Otwórz panel`, please write tests or something, because I don't want to see 3rd PR for this issue.
> 
> Additionally I think by default (if backend is not accessible etc) we should display "Zaloguj się"

so ad 1. it only happens on preview, i can't really say why, because I can't reproduce this bug on build or dev, anyway it's probably related to the fact that /auth/status whifch is fetched on navbar, doesn't exist yet on production backend, which is used by preview

ad 2. it by default displays "Zaloguj się", which again I tested on preview and dev :3

PR #634, opened 2025-07-11
chore(deps): upgrade to `nuxt` `v4`

Norbiros:


Norbiros:
Depends on [https://github.com/nitrojs/nitro/issues/3480](https://github.com/nitrojs/nitro/issues/3480).
If it takes too long to fix, we could probably use `node .output/server/index.mjs`.

Issue #489, opened 2025-04-26
[Enhancement]: Handle custom 404 error messages


Norbiros:
### My idea

Currently we override all `404` messages to our custom one, but it is not desired behaviour

PR #641, opened 2025-07-19
Enhancement: Handle custom 404 error messages

WitaLisek:
Issue #489

### DIFF
--- a/frontend/app/error.vue
+++ b/frontend/app/error.vue
@@ -13,14 +13,20 @@ onMounted(() => {
 })
 
 const errorMessage = computed(() => {
-  switch (props.error?.statusCode) {
-    case 404:
-      return 'Uwaga rycerzu,\n ta strona zniknęła jak zamek w chmurach.\n Wróć na właściwą drogę!'
-    case 500:
-      return 'Rycerz napotkał przeszkodę\n na swojej drodze.\n Spróbuj ponownie za chwilę.'
-    default:
-      return props.error?.message
+  const error = props.error
+  if (!error) {
+    return 'Nieznany błąd'
   }
+  const message = error.message?.toString().toLowerCase() || ''
+  if (error.statusCode === 404) {
+    if (message.includes('page not found')) {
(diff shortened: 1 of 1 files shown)


frontend/app/error.vue
```diff
@@ -13,14 +13,20 @@ onMounted(() => {
...
+  if (error.statusCode === 404) {
+    if (message.includes('page not found')) {
       return 'Uwaga rycerzu,\n ta strona zniknęła jak zamek w chmurach.\n Wróć na właściwą drogę!'
-    case 500:
-      return 'Rycerz napotkał przeszkodę\n na swojej drodze.\n Spróbuj ponownie za chwilę.'
-    default:
-      return props.error?.message
+    }
+  }
+  if (error.statusCode === 500) {
```
Norbiros, line 26:
```suggestion
  else if (error.statusCode === 500) {
```

frontend/app/error.vue
```diff
@@ -13,14 +13,20 @@ onMounted(() => {
...
+    return 'Rycerz napotkał przeszkodę\n Na swojej drodze.\n Spróbuj ponownie później.'
   }
+  return error.statusMessage
```
Norbiros, line 29: You probably want to return `.message`, here because now I get
image

Norbiros (changes requested):

frontend/app/error.vue
```diff
@@ -13,14 +13,20 @@ onMounted(() => {
 })
 
 const errorMessage = computed(() => {
-  switch (props.error?.statusCode) {
-    case 404:
+  const error = props.error
+  if (!error) {
+    return 'Nieznany błąd'
+  }
```
Norbiros, line 19: You don't have to change that, but we are using K&R braces style, so you could write this like:
```ts
  if (!error)
    return 'Nieznany błąd'
```
or (but I am not sure if eslint will allow for that)
```ts
  if (!error) return 'Nieznany błąd'
```

Issue #603, opened 2025-05-31
[Bug]: Whitemode is default


Norbiros:
### Description

Currently, our website by default is white (there is no `class="dark"`) in body.
This causes white flash issues when I enable a lot of cloudflare optimizations...

### Steps to reproduce

1. Install NoScript extension
2. Open our website
3. Observe the white mode

PR #651, opened 2025-07-26
Bug: Whitemode is default

WitaLisek:
Issue #603


Norbiros (approved):
Now I can turn more cloudflare optimalizations, yaay!

PR #656, opened 2025-07-28
refactor(frontend): Merge event timer page into error.vue

wojpo:


frontend/app/middleware/event-access-guard.ts
```diff
@@ -0,0 +1,28 @@
...
+            data: to,
+          })
+        case 410:
+          description = 'Nie możesz otworzyć tej strony po zakończeniu wydarzenia'
+          toast.add({ title: `Błąd ${response.code}`, description: `${description}`, color: 'error' })
+          return '/'
+      }
+    }
+  } catch {
+    toast.add({ title: 'Błąd', description: `${description}`, color: 'error' })
```
Norbiros, line 25: `useFetch` never throws an error, errors from $fetch are caught and passed to the `error` property

frontend/app/error.vue
```diff
@@ -5,8 +5,25 @@ import { setFavicon } from '~/utils/setFavicon'
 const props = defineProps({
   error: Object as () => NuxtError,
 })
-
+const showTimer = ref(false)
+const redirectPage = ref('/tasks')
 useOgImage()
```
Norbiros, line 10: Clean this up, like define are const in similar place etc

frontend/app/error.vue
```diff
@@ -34,29 +51,44 @@ const errorMessage = computed(() => {
...
       <div class="flex flex-col md:mr-10 max-w-3/4 md:max-w-3/5 space-y-5">
-        <h1 class="text-balance text-8xl text-primary font-bold">
-          {{ error?.statusCode }}
-        </h1>
-        <h2 class="whitespace-pre-line text-2xl text-white">
-          {{ errorMessage }}
-        </h2>
-        <LazyUModal title="Więcej informacji o błędzie:" hydrate-on-visible>
-          <UButton label="Więcej informacji..." variant="outline" class="w-fit mt-4" />
+        <template v-if="showTimer">
```
Norbiros, line 54: 1. I am not sure if this is good idea performance wise, because I don't think nuxt cannot treat `<templaate>` as lazy component
2. Code doesn't look really good because in one file you handle logic for normal errors and for errors with timers,

You could just create separate components for those 2 types

Norbiros: For example if this is normal 404 error you still send requests to get event status

frontend/app/error.vue
```diff
@@ -5,8 +5,25 @@ import { setFavicon } from '~/utils/setFavicon'
...
+const redirectPage = ref('/tasks')
 useOgImage()
+const { data } = await useApi('/event/info')
+
+const futureDate = computed(() => new Date(data.value?.start_date ?? Date.now()))
+function redirectToPage() {
+  navigateTo(redirectPage.value)
+}
+
+watchEffect(() => {
```
Norbiros, line 18: Xddddd, since when we declare variables outside and then use watchEffect to update them? Why not `computed`, like we already use it for the `errorMessage`?

Norbiros (approved):

PR #657, opened 2025-07-28
chore(frontend): remove admin leaderboard

Norbiros:
The current design and implementation have many issues. It will be completely redesigned for the next event, so I removed it to avoid maintaining technical debt.


WitaLisek (approved):
Very well argumentation

Issue #595, opened 2025-05-28
[Enhancement]: Cleanup `email_verification_request` table


Norbiros:
### My idea

Run code like this:

```sql
DELETE FROM email_verification_request
WHERE expiration_time < CURRENT_DATE;
```

When server starts

PR #676, opened 2025-08-14
enhancement: Delete expired email verification requests on server startup

Kubaryt:
Resolves: #595

### DIFF
--- a/backend/tests/models/email_verification_model.rs
+++ b/backend/tests/models/email_verification_model.rs
@@ -1,0 +1,91 @@
+use crate::test_utils::database::TestDatabase;
+use chrono::Utc;
+use hack4krak_backend::entities::email_verification_request;
+use sea_orm::{EntityTrait, Set};
+use std::ops::{Add, Sub};
+use uuid::Uuid;
+
+#[actix_web::test]
+async fn delete_expired_email_verification_requests() {
+    let database = TestDatabase::new().await;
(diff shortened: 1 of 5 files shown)


backend/tests/models/email_verification_model.rs
```diff
@@ -0,0 +1,65 @@
...
+use std::ops::{Add, Sub};
+use uuid::Uuid;
+
+#[actix_web::test]
+async fn delete_expired_email_verification_requests() {
+    let database = TestDatabase::new().await;
+    let id_expired = Uuid::new_v4();
+
+    email_verification_request::Entity::insert(email_verification_request::ActiveModel {
+        id: Set(id_expired),
```
Norbiros, line 14: Add also one with null and check if it is not removed

Norbiros (changes requested):

backend/tests/models/email_verification_model.rs
Kubaryt: WHAT THE F*CK ARE YOU TALKING ABOUT NULL DOESN'T EXIST IT IS JUST YOUR FREAKING DELUSION, THERE IS NOT A THING CALLED NULL

Kubaryt: (/s)

Norbiros (approved):

Issue #583, opened 2025-05-25
[Enhancement]: Add profiles to `docker-compose.yaml`


Norbiros:
### My idea

https://docs.docker.com/compose/how-tos/profiles/

Kubaryt:
@Oskiks are you still working on that? if not I will take over this

PR #684, opened 2025-08-18
feat: Add backend and frontend services to docker compose and implement profiling (#583)

Kubaryt:
Resolves: #583


DEPLOYMENT.md
```diff
@@ -59,4 +59,42 @@ the frontend app.
 
 - `NUXT_SITE_URL` - Frontend website URL
 - `BACKEND_ADDRESS` - address of the backend server (accessible from the frontend server and global network)
-- `NUXT_ERROR_API_TOKEN` - token used to report errors to the nuxterror.com
\ No newline at end of file
+- `NUXT_ERROR_API_TOKEN` - token used to report errors to the nuxterror.com
+
+## 🐋 Docker compose
+
+We have a [docker-compose.yml](docker-compose.yml), which has implemented profiles, used to start a production servers.
```
Norbiros, line 66: do not use "we", this looks nice

https://chatgpt.com/s/t_68a425c44e248191807a247cb9c7bee7

docker-compose.yml
```diff
@@ -12,6 +12,35 @@ services:
       - "5432:5432"
     volumes:
       - database:/var/lib/postgresql/data
+    profiles:
+      - db
+      - all
+
+  backend:
+    container_name: hack4krak_backend
```
Norbiros, line 20: You could add some default env variables here so it works more or less if you don't have full .env

BACKEND_ADDRESS: ${BACKEND_ADDRESS:-http://backend:8080}

Norbiros (changes requested):

docker-compose.yml
Kubaryt: funny thing is http://backend:8080 actually doesn't work and you need 0.0.0.0:8080 if you want to access api locally xdd

Norbiros: cuz we don't have an internal docker network and you forward all ports to the host

Norbiros (dismissed):
but I didn't check the compose file

Issue #628, opened 2025-06-26
[Enhancement]: Send mass emails with bcc


Sanielia:
### What to refactor?

Right now mass emails are sent with users emails set as normal receivers which causes them to see eachother emails and we could violate RODO with this 💀💀💀

Norbiros:
Refactor is clean up of the code. This is an enhancement, I created that category exactly for that purpose. In future I'll release `@norbiros/issue-templates` v2 with improved project organization

PR #685, opened 2025-08-18
enhancement: Send emails with bcc (#628)

Kubaryt:
Resolves: #628

### DIFF
--- a/backend/src/services/emails.rs
+++ b/backend/src/services/emails.rs
@@ -3,6 +3,7 @@ use crate::routes::admin::EmailSendingModel;
 use crate::utils::app_state::AppState;
 use crate::utils::error::Error;
 use actix_web::HttpResponse;
+use lettre::message::header::{Header, MailboxesHeader};
 use lettre::message::{Attachment, Mailbox, Mailboxes, MultiPart, SinglePart, header};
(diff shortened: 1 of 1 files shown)


Kubaryt:
> This is a bug, so write regression tests

there is a problem with writing this as e2e tests, our smtp test client doesn't distinguish between To and Bcc headers, so best I can do is test if recipients header is of type bcc or to

backend/src/services/emails.rs
```diff
@@ -91,15 +92,19 @@ impl Email {
             email_body = email_body.singlepart(logo_attachment);
         }
 
+        let mut has_multiple_recipients = true;
```
Norbiros, line 95: I do not like this idea it is undocumented and unlogical. IMO just leave one argument for adress and a new one for bcc recipients

Norbiros: If you really want you can drop this PR and I'll visit it in my emails refactor

Norbiros (changes requested):

backend/src/services/emails.rs
Kubaryt: maaaaybe i will do just that

Issue #663, opened 2025-08-09
[Enhancement]: Pin to specific rust version


Norbiros:
### My idea

Today we again had a situation where a random PR broke because of breaking changes in the new rustfmt/clippy update.

To resolve this, we could create a rust toolchain file, but it would need frequent updates.

Related to https://github.com/dependabot/dependabot-core/issues/1702

PR #691, opened 2025-08-27
enhancement: Pin rust to specific version and enable rust updates in renovate

Kubaryt:
Resolves: #663


Norbiros (changes requested):
> If a [toolchain file](https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file) (i.e., rust-toolchain or rust-toolchain.toml) is found in the root of the repository and no toolchain value is provided, all items specified in the toolchain file will be installed.

https://github.com/actions-rust-lang/setup-rust-toolchain/issues/47 the task is marked as good first issue, so you can look at it

PR #704, opened 2025-09-22
refactor(frontend): clean up git metadata retrieval logic

Norbiros:
Nuxt shouldn't be responsible for verifying what commit it is. I moved that logic to `Dockerfile`, so in frontend we can access env variables directly. This commit should also resolve some build issues related to copying the entire `.git` directory to `app/` directly.


Norbiros:
Merging, because it resolves deployment issues

Issue #717, opened 2025-10-18
[Bug]: Log out button logs out but doesn't refresh the site


goteusz-maszyk:
### Description

When clicking log out button the session is destroyed, but the website still shows the panel link and considers the user to be logged in. You cannot however open some links and get weird redirects until you hit ctrl+r

### Steps to reproduce

1. Log in (i used via github)
2. Moje konto
3. Wyloguj się
4. While the tokens are removed the frontend thinks you are logged in for a while and you can still open the panel

PR #257, opened 2025-02-23
Enhancement/fixes on teams page

Kubaryt:
resolves: #254 

TODO
- [x]  add responsivnes :3 @Kubaryt

PR #718, opened 2025-10-18
fix: Refresh Nuxt data before redirect, when making changes to cached data

Sanielia:
Resolves: #717

### DIFF
--- a/frontend/app/components/panel/modal/ConfirmDeleteModal.vue
+++ b/frontend/app/components/panel/modal/ConfirmDeleteModal.vue
@@ -29,7 +29,7 @@ async function onSubmit() {
   open.value = false
 
   await refreshNuxtData()
-  navigateTo(props.redirectTo)
+  await navigateTo(props.redirectTo)
 }
 </script>
 


Kubaryt:
> > > Have you tried using [`refreshNuxtData`](https://nuxt.com/docs/4.x/api/utils/refresh-nuxt-data)?
> > 
> > 
> > yeah it works, but we used external: true in deleting account, so i figured we use it for some reason instead of refreshNuxtData
> 
> There are 2 use cases of `{ external: true }`. One of them is inside `InviteUser.vue` component. It was added as a part of [4c06ebd](https://github.com/Hack4Krak/Hack4KrakSite/commit/4c06ebd6df34933a9e36f917da41a5fddfce7993) (#257).
> 
> In that PR I wrote 2 comments: image
> 
> Why it was merged? I have no idea :-/ While you are at it, please remove all other uses of `external: true`

YAY I LOVE ADDITIONAL WORK <3333

Norbiros (approved):

PR #740, opened 2025-11-17
chore: Regenerate bun lockfile

Sanielia:
Currently our bun.lock is generated with bun version 1.3.2, I base this information on fact that in line 3 there is `configVersion: 0`, whereas configVersion property was added in bun 1.3.1.

If I run bun install on master it removes this property, however it causes problems with building (it cannot resolve vue while importing it). Running it with frozen-lockfile doesn't remove the configVersion, runs properly without errors and doesn't pose any problems when building, however on #731 even running it with frozen-lockfile messes up build. Regenerating bun lockfile from scratch fixes this problem.

What should also be done in this pr is assuring that renovate uses same bun version as this project when updating deps, therefore this is WIP

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -1,6 +1,5 @@
 {
   "lockfileVersion": 1,
-  "configVersion": 0,
   "workspaces": {
     "": {
       "name": "hack4krak",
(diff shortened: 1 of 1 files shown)


Kubaryt:
as bun update has been merge it is no longer required, also #744 has been created

PR #750, opened 2025-12-12
fix: Add missing alt text to satisfy Bing requirements for indexing

Sanielia:


Kubaryt:
> Those `alt`s were missing for a reason. If you really need them use empty atrtributes, but I'm not sure if it will help. 

we were using empty attributes and bing didn't like that

> Do we have a page that has no images with missing alt tags, that you could verify?
`/tasks` has missing tag only in navbar, while it has a lot of images which have alt tag and none of them get flagged

Norbiros:
> we were using empty attributes and bing didn't like that

Are you sure about that? It is a very common practice to include empty `alt` tags for decorative images.

Issue #665, opened 2025-08-09
[Feat]: Navbar redo


wikipop:
### My suggestion

Make navbar comply with designs

Norbiros:
> For now we will omit logedin/loged out functionality as it requires proper user session storage from what I was informed by @Norbiros 

We do not need that. That is optimization concern, currently we send request to check if you are logged in, so there is no problem to also get username.

Also we have a lot of free time, so instead of implementing it partially, properly mark that this issue depends on others, so we can ship finished features.

PR #755, opened 2025-12-22
feat: Display username in Navbar

Sanielia:
Resolves: #665


Sanielia:
eeeeehhhh it will wait in slumber for auth refactor

PR #759, opened 2026-01-11
enhancement(frontend): Make about us compliant with new informations

wojpo:


frontend/content/about-us-timeline.ts
```diff
@@ -13,27 +13,27 @@ export const aboutUsTimeline: EventCardProps[] = [
...
       + 'Wydarzenie pozwoliło również na przetestowanie i potwierdzenie działania naszej autorskiej platformy.',
-    participants: 67,
+    participants: '60+',
     img: '/img/events/event1.webp',
   },
   {
     title: 'Hack4Krak CTF dla uczniów szkół podstawowych',
     subtitle: 'Maj 2025',
-    description: '**123 uczniów** szkół podstawowych z Krakowa wzięło udział w drugiej edycji Hack4Krak CTF. \n\n'
+    description: 'Ponad **123 uczniów** szkół podstawowych z Krakowa wzięło udział w drugiej edycji Hack4Krak CTF. \n\n'
```
Norbiros: For the first ctf you replaced `67` with `60` 2 times, but here you left `123` in one place. I don't think it was intended. IMO we could display the `<smth>+` numbers for `participants`, but leave the values in description as they were.

Issue #535, opened 2025-05-05
[Enhancement]: Add frontend/content to dev watch


wikipop:
### My idea

Self explanatory

Norbiros:
While working on this issue, please, revisit the idea of `ServerRenderer`. At this moment, the website uses an undefined behavior:
1. Browser sends a request to `api.hack4krak.pl`
2. It receives the original markdown for a task
3. That markdown is sent to `hack4krak.pl` (in server component render request)
4. Markdown is then returned and rendered

It could be improved in multiple ways:
- make the frontend send the request for the original markdown
- cache rendered markdown - identically to `@nuxt/content` implementation
- verify and support advanced `@nuxt/content` rendering - for example code text rendering (this should be solved as of Nuxt UI v4, but I am unsure)

PR #787, opened 2026-02-21
fix(frontend): fix copy button not working in code blocks

### LINKED ISSUE #786: Missing copy icon in code blocks
Copy icon is not displayed in code blocks.

Image

lajczi:
Fixes copy button in code blocks not working inside server components by attaching native click listeners via a client plugin with MutationObserver. Probably horror implementation

Closes #786

### DIFF
--- a/frontend/app/plugins/code-block-copy.client.ts
+++ b/frontend/app/plugins/code-block-copy.client.ts
@@ -19,7 +19,9 @@ export default defineNuxtPlugin(() => {
       navigator.clipboard.writeText(pre.textContent ?? '')
       clearTimeout(timer)
       btn.innerHTML = checkSvg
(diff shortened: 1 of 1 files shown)


Norbiros (changes requested):
I appreciate your PR, but we won’t be merging it in its current state xdddddd

I’m not entirely sure what the best fix is right now. One option could be using [a combination of client and server components](https://nuxt.com/docs/4.x/directory-structure/app/components#paired-with-a-client-component), though I haven’t tried that myself.

Even if that works, it would only be a temporary hotfix. The current approach with `MarkdownContent.server.vue` has way, way more issues, and the system really needs a full redesign. For example the code block is not the only intractable component that can be used in the task description. 

 I plan to create a proper solution in https://github.com/Hack4Krak/Hack4KrakSite/issues/535.

PR #791, opened 2026-02-25
feat(frontend): add padding to tasks page

### LINKED ISSUE #546: [Enhancement]: Add padding to tasks map
### My idea

Now map takes almost 100% width and height of site.
It would look better to have p-4 or smth around this map.

lajczi:
Closes #546

I also added `select-none` to prevent this from happening:
image

Before:
image

After:
image


Sanielia:
it ain't chore, I would call this a feature

lajczi:
> Could you please provide reproduction steps for this issue with selection, that you are fixing?

https://github.com/user-attachments/assets/ba2a0690-7e3d-4316-b32c-789ff10784b9

PR #800, opened 2026-03-01
feat: add callback url support across pages

### LINKED ISSUE #455: [Enhancement]: Callback link from /login
### My idea

Auto redirect user to destination he was previously trying to access.
e.g.
user -> /panel/join (401) -> /login?callback=/panel/join (201) -> /panel/join

lajczi:
Closes #455


Sanielia (dismissed):

Norbiros (changes requested):
Ideally it would be great to pass it to the backend while registering, include it in registration confirm link, but I don't think anybody cares about that, so just fix the issues mentioned below

Norbiros (changes requested):

Sanielia (changes requested):
there is no callback on account/submit_personal_information

Sanielia (dismissed):

frontend/app/middleware/auth.global.ts
```diff
@@ -5,15 +5,15 @@ export default defineNuxtRouteMiddleware(async (to) => {
         redirect: 'error',
       })
       if (error.value || !data.value) {
-        return '/login'
+        return await navigateTo({ name: 'login', query: { callback: to.fullPath } })
       }
       if (to.path.startsWith('/panel')) {
         if (data.value.has_personal_information === false) {
-          return '/account/submit_personal_info'
+          return { path: '/account/submit_personal_info', query: { callback: to.fullPath } }
```
Norbiros: Once you are using `navigateTo`, once just ` { path: '/account/submit_personal_info`. Is there any specific reason why?

backend/src/routes/auth/register.rs
```diff
@@ -18,6 +18,8 @@ pub struct RegisterModel {
     pub email: String,
     #[validate(length(min = 8, max = 32))]
     pub password: Password,
+    #[validate(length(max = 256))]
+    pub callback: Option<String>,
```
Norbiros: I want to be a bit more safe that there is no security issue and I want to be super, super safe. Add  a simple verification to check if callback starts with `/`, so there is no way anybody would redirect users to their own domain.

Norbiros (changes requested):
I'm a bit eppy, so I might be hallucinating/not seeing something, but how is that even supposed to work?

Sanielia (changes requested):
Also add callback coming back to /panel/profile in frontend/app/panel/profile.vue in Zmień lub zobacz informacje o końcie
```vue
          <UButton icon="mdi:account" variant="ghost" color="neutral" @click="navigateTo('/account/submit_personal_info')">
            Zmień lub zobacz informacje o koncie
          </UButton>
```

Sanielia (dismissed):

Norbiros (changes requested):
I would love to see some backend tests for this callback url, to test if callback_url is really persisted between first request, and then url in the email.

lajczi:
Long-running PR that has come a long way — the scope expanded from a simple login redirect to full callback URL support across login, registration, email confirmation, and `submit_personal_info`. The backend implementation (persisting the callback URL through the registration email flow) is the tricky part and it's good to see that addressed.

The backend test coverage added in the latest commit directly addresses Norbiros's last review comment — verifying that the callback URL is correctly persisted and survives the email confirmation round-trip is exactly the test that matters here.

The max-length validation added to the callback parameter is a nice defensive touch. One thing worth checking: is the callback URL validated to be a relative path or same-origin URL? Accepting arbitrary external URLs as a callback could enable open redirect attacks where a crafted link redirects users to a phishing page after login.

Looks close to mergeable — awaiting Norbiros's re-review.

---

backend/src/utils/validation.rs
```diff
@@ -0,0 +1,8 @@
+use validator::ValidationError;
+
+pub fn validate_callback(callback: &str) -> Result<(), ValidationError> {
+    if callback.starts_with('/') {
+        return Ok(());
+    }
+    Err(ValidationError::new("invalid_callback"))
```
Sanielia:
```suggestion
    Err(ValidationError::new("invalid_callback")).with_message("some cool info about the error")
```

backend/tests/routes/auth.rs
```diff
@@ -245,3 +245,163 @@ async fn reset_password_flow() {
...
+        .unwrap()
+        .to_string();
+    assert!(
+        refresh_header.contains("callback=%2Ftasks"),
+        "Redirect should contain callback parameter, got: {refresh_header}"
+    );
+}
+
+#[actix_web::test]
+async fn email_confirmation_without_callback() {
```
Sanielia: just move this check to email_confirmation_success, there is no need to create whole entire test case which is just duplication of that one

frontend/app/components/AuthForm.vue
```diff
@@ -18,6 +18,7 @@ const toast = useToast()
 const OAuthBaseUrl = `${useRuntimeConfig().public.openFetch.api.baseURL}/auth/oauth`
 
 const route = useRoute()
+const callback = route.query.callback as string | undefined
```
Sanielia: are you sure this cast is safe?

Sanielia (changes requested):

frontend/app/components/AuthForm.vue
Sanielia: there are more casts like that btw

lajczi: oh really?

frontend/app/middleware/guest.ts
```diff
@@ -1,10 +1,10 @@
-export default defineNuxtRouteMiddleware(async () => {
+export default defineNuxtRouteMiddleware(async (to) => {
   try {
     const user = await useAuth('/account/', {
       redirect: 'error',
     })
     if (!user.error.value && user.data.value) {
-      return '/panel'
+      return (to.query.callback as string) || '/panel'
```
lajczi: do you mean this? @Sanielia

Sanielia: yup

backend/tests/routes/auth.rs
```diff
@@ -245,3 +257,116 @@ async fn reset_password_flow() {
...
+        .to_str()
+        .unwrap()
+        .to_string();
+    assert!(
+        refresh_header.contains("callback=%2Ftasks"),
+        "Redirect should contain callback parameter, got: {refresh_header}"
+    );
+}
+
+#[cfg(feature = "full-test-suite")]
```
Sanielia, line 297: i don't think it requires sending email here, i already add mocking system for smtp service in #836 so just add comment under this pr about this tests case so i can refactor it in that commit

Sanielia (dismissed):

Sanielia (approved):

PR #822, opened 2026-03-21
feat(frontend): redesign Og Image

### LINKED ISSUE #559: [Enhancement]: Redesing OgImage
### My idea

It includes some smaller changes like:
- try improving performance (caching etc.)
- change color to mach our ui primary

lajczi:
Closes #559


Norbiros:
Probably depends on https://github.com/nuxt-modules/og-image/pull/548

frontend/public/img/logo-dark.png
Norbiros, line 1: Do not change this name, we use `/img/logo.png` in multiple places like emails

frontend/nuxt.config.ts
```diff
@@ -14,9 +14,9 @@ export default defineNuxtConfig({
     '@nuxt/ui',
     '@nuxt/eslint',
     '@nuxt/image',
+    '@nuxtjs/seo',
```
Norbiros, line 17: Why did you reorder this?

Norbiros (changes requested):
Please make the CI pass and finish the PR

frontend/nuxt.config.ts
lajczi: image

https://nuxtseo.com/docs/nuxt-seo/migration-guide/v4-to-v5#ensure-correct-module-order

Norbiros:
For now, you can implement a minimal version, that just updates the version, uses proper colours and in the future you can try redesigning and making the OgImage even better

frontend/nuxt.config.ts
Norbiros: Could you please add that comment? I am sure that without it someone will accidentally reorder it, break a lot of random stuff and spend 1 hour debugging it.

lajczi: sure

lajczi:
Good call on the minimal-first approach suggested by Norbiros — getting the OG image using correct colours and the proper module version into production sooner is more valuable than waiting for a perfect redesign.

The module ordering change in `nuxt.config.ts` is a good catch — adding an inline comment explaining that `@nuxt/ui` must come before `@nuxtjs/seo` per the v4→v5 migration guide will save future developers from a frustrating debugging session. Norbiros confirmed this is needed and you already agreed to add it — just making sure it's in the final commit.

Once the CI passes and that comment is in place this looks ready to review properly.

---

lajczi:
eh still not ready

lajczi:
this will also be recreated

PR #842, opened 2026-04-06
feat: QR code based identity verification and tags application

### LINKED ISSUE #498: [Feat]: On premise team verification
### My suggestion

Extend the QR-code identity system with either:
- a new action for confirming member's team
- or just when confirming user attendance, update their team status

### LINKED ISSUE #836: [Feat]: QR-code based identity verification
### My suggestion

During the event implement a system, where participants can showcase their QR code "ticket"/passport to the staff to confirm their identity. This should be combined with system of attributes, so admins can mark users after scanning (fe. registered, received dinner 1). The idea was discussed during our design sprint, *so there is no need to describe it in detail here.*

Note: The frontend doesn't have to be perfect - it will be adjusted when redesigning the user panel, but try to do your best while implementing user & admin components.

Sanielia:
Resolves: #836 
Resolves: #498 

Todo:
- [x] implement team verification based on this verification system #498


frontend/app/pages/panel/profile.vue
```diff
@@ -53,6 +78,9 @@ async function logout() {
           <UButton icon="mdi:account-cog" variant="ghost" color="neutral" @click="updateAccountModal = true">
             Zmień ustawienia konta
           </UButton>
+          <UButton icon="mdi:qrcode" variant="ghost" color="neutral" @click="qrCodeModal = true">
+            Zobacz swój identyfikacji kod QR
```
Norbiros: Poprawna Polski była ten

frontend/app/pages/admin/verification.vue
```diff
@@ -0,0 +1,264 @@
...
+            variant="outline"
+            @click="cancelTagApplication"
+          >
+            Anuluj
+          </UButton>
+        </div>
+      </div>
+    </div>
+
+    <!-- QR Scanner Modal -->
```
Norbiros: noop

frontend/app/pages/admin/verification.vue
```diff
@@ -0,0 +1,264 @@
...
+    if (mode.value === 'identify') {
+      const response = await $api('/admin/verification/identify', {
+        method: 'POST',
+        body: {
+          verification_uuid: verificationUuid.value,
+        },
+        credentials: 'include',
+      })
+
+      userInfo.value = response as typeof userInfo.value
```
Norbiros: :cat:

frontend/app/pages/admin/verification.vue
```diff
@@ -0,0 +1,264 @@
...
+  isLoading.value = true
+
+  try {
+    await $api('/admin/verification/confirm-tag', {
+      method: 'POST',
+      body: {
+        user_id: userInfo.value.user_id,
+        tag_id: pendingTag.value,
+      },
+      credentials: 'include',
```
Norbiros: $auth

backend/Cargo.toml
```diff
@@ -48,6 +48,7 @@ tokio.workspace = true
 # Utils
 regex = "1.12.3"
 futures-util = "0.3.32"
+async-trait = "0.1"
```
Norbiros: nope

backend/tests/test_utils/verification_mock.rs
```diff
@@ -0,0 +1,20 @@
+use async_trait::async_trait;
+use hack4krak_backend::services::verification::VerificationQrCodeSender;
+use hack4krak_backend::utils::app_state::AppState;
+use hack4krak_backend::utils::error::Error;
+use uuid::Uuid;
+
+mockall::mock! {
```
Norbiros: Maybe automock, and use proper design pattern for changing mocking emails using DI

Norbiros (changes requested):

backend/tests/mocks/smtp_mock.rs
Norbiros, line 1: As this is our first real mock, maybe we could create a `tests/mocks/` directory for them?

backend/src/entities/sea_orm_active_enums.rs
```diff
@@ -1,16 +1,8 @@
-//! `SeaORM` Entity, @generated by sea-orm-codegen 1.1.11
+//! `SeaORM` Entity, @generated by sea-orm-codegen 1.1.20
 
 use sea_orm::entity::prelude::*;
 use serde::{Deserialize, Serialize};
 
-#[derive(Debug, Clone, PartialEq, Eq, EnumIter, DeriveActiveEnum, Serialize, Deserialize)]
-#[sea_orm(rs_type = "String", db_type = "Enum", enum_name = "team_status")]
-pub enum TeamStatus {
```
Norbiros, line 8: Okay, but how can we query if team has been confirmed? Do we have to loop over all members to check if the team is taking part in our competition? How also can we possibly ban teams? I think that we should keep this field

backend/src/models/user_participant_tags.rs
```diff
@@ -0,0 +1,119 @@
...
+        active_model.save(database_connection).await?;
+
+        Ok(())
+    }
+
+    pub async fn get_or_create(
+        database_connection: &DatabaseConnection,
+        user_id: Uuid,
+    ) -> Result<Model, Error> {
+        // Used in tests
```
Norbiros, line 66: Then testing this code is missing a point. I would prefer to mock it or spin up a postrgres db for testing. BTW, why it doesn't work in sqlite?

backend/templates/email/verification_qr_code.html
```diff
@@ -0,0 +1,27 @@
...
+{% block content %}
+<h1 style="margin: 0 0 24px; font-size: 24px; line-height: 32px; font-weight: 600; color: #0f172a">
+    Cześć, {{ user }}!
+</h1>
+<p style="margin: 0 0 24px; font-size: 16px; line-height: 24px; color: #475569">
+    Poniżej znajduje się Twój unikalny kod weryfikacyjny. Pokaż go organizatorom wydarzenia, aby potwierdzić swoją tożsamość.
+</p>
+
+<div style="text-align: center; margin: 24px 0;">
+    <img src="https://api.qrserver.com/v1/create-qr-code/?size=200x200&data={{ verification_id }}"
```
Norbiros, line 14: I am not sure if leaking all qrcodes to external service is a good idea. I remember that we had a system of attachements but I am not sure if we still have it.

backend/src/services/verification.rs
```diff
@@ -0,0 +1,164 @@
...
+        })
+    }
+
+    pub async fn reset_verification_id(app_state: &AppState, user_id: Uuid) -> Result<(), Error> {
+        let user = users::Entity::find_by_id(user_id)
+            .one(&app_state.database)
+            .await?
+            .ok_or(Error::UserNotFound)?;
+
+        let username = user.username.clone();
```
Norbiros, line 145: Why are you cloning username?

backend/src/services/verification.rs
```diff
@@ -0,0 +1,164 @@
...
+        let user_email = user.email.clone();
```
Norbiros, line 146: Same here

backend/src/services/verification.rs
```diff
@@ -0,0 +1,164 @@
...
+#[derive(Serialize, Deserialize, ToSchema, Debug)]
+pub struct UserVerificationInfo {
+    pub verification_id: Uuid,
+    pub applied_tags: Vec<ParticipantTag>,
+}
+
+pub struct VerificationService;
+
+impl VerificationService {
+    pub async fn send_verification_qr_email(
```
Norbiros, line 30: There is no need to mention qr here

backend/src/routes/flag/submit.rs
```diff
@@ -54,6 +57,17 @@ pub async fn submit(
     let event_config = app_state.task_manager.event_config.read().await;
 
     if !event_config.is_after_event() {
+        let user_participant_tags =
+            user_participant_tags::Model::get_or_create(&app_state.database, user.id).await?;
+        let has_verified_tag = user_participant_tags.has_tag_type(
+            &*app_state.task_manager.participant_tags_config.read().await,
+            "verified",
```
Norbiros, line 64: I would prefer to have defined consts in proper file

backend/src/routes/flag/submit.rs
```diff
@@ -54,6 +57,17 @@ pub async fn submit(
...
+        );
+
+        if !has_verified_tag {
+            return Err(Flag(FlagError::ParticipantNotVerified));
```
Norbiros, line 68: Isn't this middleware's responsibility for verification of that?

backend/src/routes/admin/verification/reset_uuid.rs
```diff
@@ -0,0 +1,26 @@
...
+    responses(
+        (status = 200, description = "Verification UUID successfully reset and email sent."),
+        (status = 404, description = "User not found."),
+        (status = 500, description = "Internal server error.")
+    ),
+    operation_id = "admin_verification_reset_uuid",
+    tag = "admin/verification"
+)]
+#[post("/reset/{user_id}")]
+pub async fn reset_uuid(
```
Norbiros, line 19: I know that I am saying this a lot, but imagine a random developer looking at the codebase, what does it mean to "reset uuid"? It also raises the second argument, that you shouldn't hardcode types in function/variable names - you are storing **verification code** and you don't care if that is uuid or a silly joke about path in the woods.

backend/src/services/verification.rs
```diff
@@ -0,0 +1,164 @@
...
+        })
+    }
+
+    pub async fn apply_tag(
+        database: &DatabaseConnection,
+        task_manager: &TaskManager,
+        verification_id: Uuid,
+        tag_id: &str,
+    ) -> Result<(), Error> {
+        let participant_tags_config = task_manager.participant_tags_config.read().await;
```
Norbiros, line 76: It won't hurt to extract this chunk (verifying code) to a separate function and it will help us to avoid code duplication in the future

backend/src/services/verification.rs
```diff
@@ -0,0 +1,164 @@
...
+    ) -> Result<UserVerificationInfo, Error> {
+        let applied_tags = Self::user_tags(database, task_manager, user.id).await?;
+
+        Ok(UserVerificationInfo {
+            verification_id: user.verification_id,
+            applied_tags,
+        })
+    }
+
+    pub async fn reset_verification_id(app_state: &AppState, user_id: Uuid) -> Result<(), Error> {
```
Norbiros, line 139: I have already mentioned that, but sometimes you use `verification id` sometimes `uuid`. I'd pick `verification_code` cuz it is more informative for me than `id`, but you can decide - just be consise

Norbiros (changes requested):
That's not complete - I ran out of time, but this should be something to work on for now. Ping me if anything in the comments needs clarification.

backend/src/models/user_participant_tags.rs
Sanielia: sqlite doesn't have arrays

backend/src/services/verification.rs
Sanielia: i move user later into UpdatableModel::update

backend/src/services/verification.rs
Sanielia: i move user later into UpdatableModel::update

backend/src/services/verification.rs
Sanielia: there absolutely is, otherwise one may assume we are sending this email to verify user, which doesn't happen here

backend/src/routes/flag/submit.rs
Sanielia: which middleware

backend/src/services/verification.rs
Sanielia: i replaced verification_uuid with verification_id everywhere before your review (it seems that almost everywhere, cuz i missed spot in which verification isn't even mentioned), i think id is better, cuz verification code sounds like its something for user to submit and make them verified on website, identification_code is possibly better, and then identification not verifycation servie, cuz this is more of identification not verification, idk

Sanielia:
Apparently it requires redesigning like half of things, so it will take some more time to finish

lajczi:
Ambitious feature — QR-code-based on-premise identity verification is a solid approach for an event like Hack4Krak. The refactoring work (renaming to `IdentificationService`, splitting `AuthService` into `Authentication`/`Authorization`, introducing the `SMTPClient` trait) is well-structured and improves testability.

A few observations on the open discussion points:

**External QR service**: The concern about using `api.qrserver.com` is valid from both a privacy and reliability perspective. Sending verification IDs to a third party is a risk, and it also creates a hard external dependency in email flows. Generating the QR code server-side (e.g., with a Rust QR library like `qrcode`) and embedding it as an inline image or attachment would eliminate both issues.

**Team status**: Norbiros's question about how to check team confirmation state (and how to ban teams) without `TeamStatus` is worth answering clearly in the PR. If participant-level tags now encode team eligibility, the logic for querying team status needs to be documented or exposed via a dedicated query.

**Testing with Postgres arrays**: The SQLite limitation is a real constraint for tests — documenting it clearly (or moving the relevant tests to the Postgres-backed integration test setup) is better than leaving a `// Used in tests` comment that makes the test misleading.

**Naming**: The `IdentificationService` rename is the right call — it better describes what the service actually does (scan QR → identify person) vs. "verification" which reads more like credential checking.

Looking like solid progress — the draft status makes sense given the remaining open points.

---

backend/locales/error.yaml
```diff
@@ -33,6 +33,10 @@ Error:
   RecipientNotFound: "Nie ma użytkownika o nazwie: {username}"
   AccessDuringStage: "Nie możesz uzyskać dostępu do tego punktu w trakcie etapu"
   FailedToParseStage: "Nie udało się znaleźć etapu o identyfikatorze {stage_identifier}"
+  InvalidIdentificationId: "Nieprawidłowy kod identyfikacyjny UUID"
```
Norbiros: This error is facing the user, do not mention UUID

frontend/app/pages/panel/profile.vue
```diff
@@ -42,6 +45,28 @@ async function logout() {
         </h3>
         <span>Przesłane flagi: 0</span>
       </div>
+
+      <!-- Applied Tags Section -->
+      <div v-if="identification?.applied_tags && identification.applied_tags.length > 0" class="p-5">
+        <h1 class="font-bold text-xl mb-3">
```
Norbiros: Why are you using random widths like 5, 3. I am not sure if this aligned without our design docs

backend/package.json
```diff
@@ -1,4 +1,4 @@
-{
+ {
```
Norbiros: :x

frontend/app/pages/panel/profile.vue
Sanielia: istg…

Sanielia: well its not exactly aligned with design doc, however those are not random values, they are the same as values from buttons section of profile panel, so maybe they all should be refactored, but i think its better we keep it consistent and i don't want to refactor panel in this pr

backend/src/services/verification.rs
Norbiros: So pick a better name - use `verification_code` or completly rewrite it

Norbiros (approved):

PR #845, opened 2026-04-07
feat(frontend): add username & context menu to navigation

### LINKED ISSUE #665: [Feat]: Navbar redo

lajczi:
Logging out was taking me a long time, so I thought about adding a context menu here

Preview:
image

Btw this PR also closes #665

### DIFF
--- a/frontend/app/components/Navbar.vue
+++ b/frontend/app/components/Navbar.vue
@@ -1,7 +1,7 @@
 <script setup lang="ts">
 import { NAVBAR_ITEMS } from '~~/content/navbar'
 
-const { data: isLoggedIn } = useAuth('/auth/status', {
+const { data: user } = useAuth('/account/', {
   redirect: 'error',
   onResponseError: undefined,
 })
@@ -18,6 +18,35 @@ const navigationMenuProperties = computed(() => ({
     list: 'gap-8',
   },
 }))
+
+const userMenuItems = computed(() => [
+  [
+    {
+      label: user.value?.username ?? 'Użytkownik',
+      type: 'label' as const,
+    },
+  ],
+  [
+    {
+      label: 'Panel',
+      icon: 'pixelarticons:dashboard',
+      onSelect: () => navigateTo('/panel'),
+    },
+    {
+      label: 'Profil',
+      icon: 'pixelarticons:user',
+      onSelect: () => navigateTo('/panel/profile'),
+    },
+  ],
+  [
+    {
+      label: 'Wyloguj się',
+      icon: 'pixelarticons:logout',
+      color: 'error' as const,
+      onSelect: () => useSession().logout(),
+    },
+  ],
+])
 </script>
 
 <template>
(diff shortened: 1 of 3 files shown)


goteusz-maszyk (dismissed):

Norbiros (changes requested):

Sanielia (commented):
Image currently if my access token expires i don't have username in navbar until full page reload which is suboptimal, it would be cool if you could find a solution to this

lajczi:
AHHHHH I broke formatting

frontend/app/components/Navbar.vue
```diff
@@ -1,11 +1,19 @@
...
 
-const { data: isLoggedIn } = useAuth('/auth/status', {
+const { $auth } = useNuxtApp()
+
+const { data: user, refresh: refreshUser } = useAuth('/account/', {
   redirect: 'error',
   onResponseError: undefined,
 })
 
+const route = useRoute()
```
Norbiros, line 11: That is not a proper solution. Problem is that probably `useAuth` doesn't probably return proper data after refreshing the access token, but I didn't look into it. Sending a new request to `/account` every time you navigate our site (even if you are not logged in) is more than suboptimal.

Norbiros (changes requested):

frontend/app/plugins/openFetch.ts
```diff
@@ -70,6 +70,10 @@ export default defineNuxtPlugin((nuxtApp) => {
           if (import.meta.server) {
             nuxtApp.ssrContext?.event.node.res.setHeader('Set-Cookie', cookies)
           }
+
+          if (import.meta.client) {
```
Norbiros, line 74: Still, I’m not sure about this fix. I see two ways you can get past the Cerberus guarding the merge of this PR (me):

1. Provide detailed information about the bug -  e.g., a video recording, steps to reproduce, and a clear explanation of the original issue. Then we can discuss whether this fix is appropriate.
2. Wait for me to review it manually :3

Norbiros (changes requested):

lajczi:
Nice feature — the context menu is a clean way to surface logout and account navigation without cluttering the navbar, and the username display closes a long-standing UX gap.

The open question around the `openFetch.ts` fix is a reasonable blocker. Calling `refreshNuxtData()` after a client-side token refresh is a reasonable pattern, but it does have side effects (all data fetched via `useAsyncData` gets re-queried), so it's worth verifying it only fires after an actual token renewal rather than on every request. A reproduction video or steps (e.g. "let token expire, navigate to a page, observe stale username, see it update correctly after refresh") would help Norbiros sign off quickly.

The Navbar component tests added in the last commit are a welcome addition — testing auth state rendering and reactive updates is exactly the kind of test that catches regressions in this flow.

---

frontend/app/plugins/openFetch.ts
lajczi: @Sanielia

Norbiros: @Sanielia provided a very helpful reproduction, and I’ve resolved the issue in #874. Could you please review / test my solution?

Norbiros:
Rebase on top of master

lajczi:
> Rebase on top of master

why? I don't see any conflict

Norbiros:
And? There were a lot of major changes on the master branch (like useApi bug fix) that you need on this branch.

Norbiros (approved):

PR #847, opened 2026-04-11
test(backend): add more tests

lajczi:
This PR adds a few more backend tests so we can bump our coverage to around ~80%


lajczi:
what i did with pr title :sob:

backend/tests/middlewares/event.rs
```diff
@@ -0,0 +1,207 @@
...
+    let tm = task_manager_with_stages(event_stages(future(1), future(2))).await;
+
+    let app = TestApp::default()
+        .with_database(test_db)
+        .with_task_manager(tm)
+        .build_app()
+        .await;
+
+    // /event/status uses EventMiddleware::disallow_before_event()
+    let request = test::TestRequest::get()
```
Sanielia: it would be better to just mock app with only this middleware and maybe custom route or smth, cuz current test would need to be changed if we just removed eventmiddleware from /event/status without changing event middleware

backend/tests/middlewares/event.rs
```diff
@@ -0,0 +1,207 @@
...
+    let test_db = TestDatabase::new().await;
+    let tm = task_manager_with_stages(event_stages(future(1), future(2))).await;
+
+    let app = TestApp::default()
+        .with_database(test_db)
+        .with_task_manager(tm)
+        .build_app()
+        .await;
+
+    // /tasks uses EventMiddleware::disallow_before_event() and /flag uses allow_only_during_event()
```
Sanielia: Image the what

backend/tests/middlewares/event.rs
Sanielia, line 1: you are just testing specific helper methods, test middleware logic directly as well

backend/tests/models/team_invites.rs
```diff
@@ -0,0 +1,282 @@
...
+use hack4krak_backend::entities::sea_orm_active_enums::TeamStatus;
+use hack4krak_backend::entities::{team_invites, teams, users};
+use hack4krak_backend::models::task::RegistrationConfig;
+use sea_orm::ActiveValue::Set;
+use sea_orm::EntityTrait;
+use uuid::Uuid;
+
+/// Helper: insert a team_invite directly with an explicit ID
+/// (needed because `invite_user` relies on PostgreSQL's `gen_random_uuid()`
+/// which is not available in the SQLite test database).
```
Sanielia:
```suggestion
```

backend/tests/models/team_invites.rs
```diff
@@ -0,0 +1,282 @@
...
+        id: Set(Uuid::new_v4()),
+        user: Set(user_id),
+        team: Set(team_id),
+    })
+    .exec(db)
+    .await
+    .unwrap();
+}
+
+// === invite_user ===
```
Sanielia:
```suggestion
```

backend/tests/models/team_invites.rs
```diff
@@ -0,0 +1,282 @@
...
+        })
+        .await;
+    let config = RegistrationConfig::default();
+
+    let result =
+        team_invites::Model::invite_user(&test_db.database, &config, user, other_team).await;
+    assert!(result.is_err());
+}
+
+// === get_invitations ===
```
Sanielia:
```suggestion
```

backend/tests/models/team_invites.rs
```diff
@@ -0,0 +1,282 @@
...
+async fn get_invitations_empty() {
+    let test_db = TestDatabase::new().await;
+    let user = test_db.with_default_user().await;
+
+    let invitations =
+        team_invites::Model::get_invitations(&test_db.database, user).await.unwrap();
+    assert!(invitations.is_empty());
+}
+
+// === get_invitations_with_teams ===
```
Sanielia:
```suggestion
```

backend/tests/models/team_invites.rs
```diff
@@ -0,0 +1,282 @@
...
+
+    let invitations =
+        team_invites::Model::get_invitations_with_teams(&test_db.database, user).await.unwrap();
+    assert_eq!(invitations.len(), 1);
+    let (_, related_team) = &invitations[0];
+    assert!(related_team.is_some());
+    assert_eq!(related_team.as_ref().unwrap().id, team.id);
+}
+
+// === accept_invitation ===
```
Sanielia:
```suggestion
```

backend/tests/models/team_invites.rs
```diff
@@ -0,0 +1,282 @@
...
+    let result = team_invites::Model::assert_user_doesnt_have_invites_from_this_team(
+        &test_db.database,
+        user,
+        team,
+    )
+    .await;
+    assert!(result.is_err());
+}
+
+// === get_invited_users ===
```
Sanielia:
```suggestion
```

backend/tests/models/team_invites.rs
```diff
@@ -0,0 +1,282 @@
...
+
+    insert_invite(&test_db.database, user.id, team.id).await;
+
+    let invited =
+        team_invites::Model::get_invited_users(&test_db.database, team).await.unwrap();
+    assert_eq!(invited.len(), 1);
+    assert_eq!(invited[0], user.username);
+}
+
+// === revoke_invitation ===
```
Sanielia:
```suggestion
```

backend/tests/models/team_invites.rs
```diff
@@ -0,0 +1,282 @@
...
+        .with_user(users::UpdatableModel {
+            team: Some(Some(team.id)),
+            ..Default::default()
+        })
+        .await;
+    let config = RegistrationConfig::default();
+
+    let result =
+        team_invites::Model::invite_user(&test_db.database, &config, user, other_team).await;
+    assert!(result.is_err());
```
Sanielia: it shouldn't be any error, but this specific error you are testing for

backend/tests/models/team_invites.rs
```diff
@@ -0,0 +1,282 @@
...
+use hack4krak_backend::entities::{team_invites, teams, users};
+use hack4krak_backend::models::task::RegistrationConfig;
+use sea_orm::ActiveValue::Set;
+use sea_orm::EntityTrait;
+use uuid::Uuid;
+
+/// Helper: insert a team_invite directly with an explicit ID
+/// (needed because `invite_user` relies on PostgreSQL's `gen_random_uuid()`
+/// which is not available in the SQLite test database).
+async fn insert_invite(db: &sea_orm::DatabaseConnection, user_id: Uuid, team_id: Uuid) {
```
Sanielia: just make it a method on testdatabase

backend/tests/models/team_invites.rs
```diff
@@ -0,0 +1,282 @@
...
+    let team = test_db.with_default_team().await;
+    let user = test_db.with_default_user().await;
+
+    insert_invite(&test_db.database, user.id, team.id).await;
+
+    let invitations =
+        team_invites::Model::get_invitations_with_teams(&test_db.database, user).await.unwrap();
+    assert_eq!(invitations.len(), 1);
+    let (_, related_team) = &invitations[0];
+    assert!(related_team.is_some());
```
Sanielia:
```suggestion
```
you are already testing if its some in unwrap below

backend/tests/models/team_invites.rs
```diff
@@ -0,0 +1,282 @@
...
+    let config = RegistrationConfig::default();
+
+    let result = team_invites::Model::accept_invitation(
+        &test_db.database,
+        &config,
+        team,
+        user,
+    )
+    .await;
+    assert!(result.is_err());
```
Sanielia: not any error

backend/tests/models/team_invites.rs
```diff
@@ -0,0 +1,282 @@
...
```
Sanielia: not any error

backend/tests/models/team_invites.rs
```diff
@@ -0,0 +1,282 @@
...
+        &test_db.database,
+        &config,
+        team2,
+        user,
+    )
+    .await;
+    assert!(result.is_err());
+}
+
+// === assert_user_doesnt_have_invites_from_this_team ===
```
Sanielia:
```suggestion
```

backend/tests/models/team_invites.rs
```diff
@@ -0,0 +1,282 @@
...
+        &config,
+        team,
+        user,
+    )
+    .await;
+    assert!(result.is_err());
+}
+
+#[tokio::test]
+async fn accept_invitation_wrong_team() {
```
Sanielia: this is just same as no invitation or smth, this exact setting should never happen

backend/tests/models/team_invites.rs
```diff
@@ -0,0 +1,282 @@
...
+
+    insert_invite(&test_db.database, user.id, team.id).await;
+
+    let result = team_invites::Model::assert_user_doesnt_have_invites_from_this_team(
+        &test_db.database,
+        user,
+        team,
+    )
+    .await;
+    assert!(result.is_err());
```
Sanielia: not any error

backend/tests/models/team_invites.rs
```diff
@@ -0,0 +1,282 @@
...
+}
+
+#[tokio::test]
+async fn revoke_invitation_user_not_found() {
+    let test_db = TestDatabase::new().await;
+    let team = test_db.with_default_team().await;
+
+    let result =
+        team_invites::Model::revoke_invitation(&test_db.database, "nonexistent", team.id).await;
+    assert!(result.is_err());
```
Sanielia: not any error

backend/tests/models/team_invites.rs
```diff
@@ -0,0 +1,282 @@
...
+
+#[tokio::test]
+async fn revoke_invitation_no_invitation() {
+    let test_db = TestDatabase::new().await;
+    let team = test_db.with_default_team().await;
+    let user = test_db.with_default_user().await;
+
+    let result =
+        team_invites::Model::revoke_invitation(&test_db.database, &user.username, team.id).await;
+    assert!(result.is_err());
```
Sanielia: not any error

Sanielia (changes requested):
you don't test like half of things:
- auth middleware
- almost none of models
Image

- half of the routes
- 70% of utils

i checkd just first 3 files

Norbiros (changes requested):
1. In Rust you have 2 ways to define tests (more info [here](https://doc.rust-lang.org/book/ch11-03-test-organization.html)):
- integration tests (in `tests/` directory)
- unit tests (using `#[cfg(test)]`)

For example your tests in `bearer.rs` are unit tests, and they shouldn't be placed in the `tests/` directory. 

2. Parametrization & code duplication. 

`tests/utils/bearer.rs` is 50 lines long, and it doesn't have it.
In other testing frameworks like pydantic we have awesome feature called parametrization (you can call one test case with different attributes). In rust we don't have something like that built-in, but there are other ways to solve it:
- find a good library that provides that feature - it needs to properly display different test cases to `cargo test`, have good dx etc (before implementing consult on your pick with me on Discord)
- just use `for` loop

3. We are not SQLite - there is no need to test every internal method. Often good and working integration tests are better - they allow us to catch more mistakes. Writing 10 tests cases that verify implementation of other library (we are just forwarding argument)  is not good use of time and adds little value. Those tests will just slow us down after each small refactor.

backend/tests/utils/task_manager.rs
```diff
@@ -0,0 +1,199 @@
+use hack4krak_backend::models::event_config::{EventConfig, EventStage, EventStageType};
+use hack4krak_backend::models::task::{TaskConfig, TaskMeta};
+use hack4krak_backend::services::task_manager::TaskManager;
+
+#[test]
+fn get_task_valid_id() {
+    let tm = TaskManager::default();
```
Norbiros, line 7:
```suggestion
    let task_manager = TaskManager::default();
```
We do not accept abbreviations

backend/tests/utils/common_responses.rs
```diff
@@ -0,0 +1,86 @@
...
+    assert!(
+        refresh_header
+            .to_str()
+            .unwrap()
+            .contains("https://example.com/callback")
+    );
+}
+
+#[test]
+fn create_redirect_response_contains_url() {
```
Norbiros, line 28: This is type of tests that I was talking about in point `3` of my review comment. There are a lot of ways to redirect user, you are just really testing Rust language rn. If we change that to a different approach, you'd have to update tests. Most of the people are lazy, and would just use AI for that.

Norbiros (changes requested):

Norbiros:
I didn't look at most of the files, because a lot of comments would be duplicated.

lajczi:
Good progress on the test coverage push — the 80% target is the right goal and the scope is ambitious. The draft status makes sense given the open feedback. A few thoughts on the review discussion:

**Integration vs unit tests placement**: Norbiros's point is spot-on for Rust. Tests that require a running context (DB, HTTP, middleware chains) belong in `tests/` as integration tests, while pure logic tests (like the bearer token parsing) are better as `#[cfg(test)]` modules in the same file as the production code. This keeps compile-time separation clear.

**Avoiding testing-library internals**: The feedback about `common_responses.rs` tests is worth taking seriously. If you're verifying that axum's redirect mechanism works, you're testing axum, not your code. Focus tests on your business logic: does the route return the right status code and body for a given input? That's more durable and more meaningful.

**Deduplication**: A simple `for` loop or helper fixture goes a long way in test files. Even without a parametrize macro, grouping similar cases into an array of `(input, expected)` pairs and iterating is idiomatic Rust and much more readable.

Overall the foundation is there — once the placement and deduplication issues are cleaned up this will be a solid PR.

---

PR #855, opened 2026-04-19
build: use Docker secrets for `GIT_ACCESS_TOKEN`

### LINKED ISSUE #854: [Enhancement]: Use Docker Secrets instead of args for providing tasks git url
### My idea

Currently we use built-time args for providing token, repository url etc:

https://github.com/Hack4Krak/Hack4KrakSite/blob/0da8cc94929eab42f78412aea85fb24647335629/backend/Dockerfile#L31-L37

This solution isn't really secure. It would be way better to use docker secrets for that

lajczi:
Closes #854

### DIFF
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -30,11 +30,12 @@ RUN apt-get update && apt-get install -y libssl3 ca-certificates git
 # Download TasksTemplate or other tasks repository
 ARG GIT_USER=Hack4Krak
 ARG GIT_REPO=TasksTemplate
-ARG GIT_ACCESS_TOKEN=token
 ARG BRANCH=master
+# Pass --build-arg CACHEBUST=$(date +%s) to invalidate cache when tasks repo updates
+ARG CACHEBUST=0
 
-ADD https://$GIT_ACCESS_TOKEN:x-oauth-basic@api.github.com/repos/$GIT_USER/$GIT_REPO/git/refs/heads/$BRANCH version.json
-RUN git clone -b $BRANCH https://$GIT_USER:$GIT_ACCESS_TOKEN@github.com/$GIT_USER/$GIT_REPO.git TasksRepo/
+RUN --mount=type=secret,id=GIT_ACCESS_TOKEN,env=GIT_ACCESS_TOKEN \
+  git clone -b $BRANCH "https://${GIT_USER}:${GIT_ACCESS_TOKEN}@github.com/${GIT_USER}/${GIT_REPO}.git" TasksRepo/
 
 ENV TASKS_BASE_PATH=TasksRepo/
 
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -21,6 +21,8 @@ services:
     build:
         context: .
         dockerfile: backend/Dockerfile
+        secrets:
+          - GIT_ACCESS_TOKEN
     restart: on-failure
     ports:
       - "8080:8080"
@@ -49,3 +51,7 @@ services:
 
 volumes:
   database:
+
+secrets:
+  GIT_ACCESS_TOKEN:
+    environment: GIT_ACCESS_TOKEN


backend/Dockerfile
```diff
@@ -25,16 +25,18 @@ RUN cargo build --release --bin hack4krak-backend
...
 
 # Download TasksTemplate or other tasks repository
 ARG GIT_USER=Hack4Krak
 ARG GIT_REPO=TasksTemplate
-ARG GIT_ACCESS_TOKEN=token
 ARG BRANCH=master
 
-ADD https://$GIT_ACCESS_TOKEN:x-oauth-basic@api.github.com/repos/$GIT_USER/$GIT_REPO/git/refs/heads/$BRANCH version.json
-RUN git clone -b $BRANCH https://$GIT_USER:$GIT_ACCESS_TOKEN@github.com/$GIT_USER/$GIT_REPO.git TasksRepo/
+RUN --mount=type=secret,id=GIT_ACCESS_TOKEN,env=GIT_ACCESS_TOKEN \
```
Norbiros: The entire point of this `ADD` command was to invalidate cache of this git repository if the head updates. Currently, both of the `git clone` and `curl` commands will be cached and that is not good behaviour.

backend/Dockerfile
```diff
@@ -10,7 +10,7 @@ COPY backend/openapi-cli/Cargo.toml ./openapi-cli/Cargo.toml
 COPY backend/rust-toolchain.toml ./rust-toolchain.toml
 # Create minimal lib.rs/main.rs stubs required by cargo
 RUN mkdir -p src macros/src migration/src openapi-cli/src \
-    && touch src/lib.rs src/main.rs macros/src/lib.rs migration/src/lib.rs openapi-cli/src/main.rs
+  && touch src/lib.rs src/main.rs macros/src/lib.rs migration/src/lib.rs openapi-cli/src/main.rs
```
Sanielia: unrelated

backend/Dockerfile
```diff
@@ -25,16 +25,18 @@ RUN cargo build --release --bin hack4krak-backend
 # We do not need the Rust toolchain to run the binary!
 FROM debian:trixie-slim AS runtime
 WORKDIR /app
-RUN apt-get update && apt-get install -y libssl3 ca-certificates git
+RUN apt-get update && apt-get install -y libssl3 ca-certificates git curl
```
Sanielia: is curl really needed? did you check if there is a way to do it without it?

docker-compose.yml
```diff
@@ -36,8 +38,8 @@ services:
   frontend:
     container_name: hack4krak_frontend
     build:
-        context: .
-        dockerfile: frontend/Dockerfile
+      context: .
+      dockerfile: frontend/Dockerfile
```
Sanielia: unrelated

Sanielia (changes requested):

docker-compose.yml
lajczi: eh it is because of my formatting settings in vscode. I will try to revert

backend/Dockerfile
lajczi: yop because `ADD` doesn't support `--mount=type=secret`

docker-compose.yml
Sanielia: do you not check what you commit?

Norbiros (changes requested):
I am not sure if the fix for this issue will be that simple. Remember that we are deploying it in dokploy, and we need to be sure that it is possible.

backend/Dockerfile
lajczi: makes sense

backend/Dockerfile
```diff
@@ -30,11 +30,12 @@ RUN apt-get update && apt-get install -y libssl3 ca-certificates git
 # Download TasksTemplate or other tasks repository
 ARG GIT_USER=Hack4Krak
 ARG GIT_REPO=TasksTemplate
-ARG GIT_ACCESS_TOKEN=token
 ARG BRANCH=master
+# Pass --build-arg CACHEBUST=$(date +%s) to invalidate cache when tasks repo updates
+ARG CACHEBUST=0
```
Norbiros, line 35: I stated that it needs to work with dokploy and there is no way to pass random parameter.

Norbiros (changes requested):

lajczi:
Great security improvement! Using Docker BuildKit secrets (`--mount=type=secret`) is the right approach for handling sensitive credentials like `GIT_ACCESS_TOKEN` — it avoids exposing the token in image layers or build history, unlike `ARG`.

A couple of observations:
- `ARG CACHEBUST=0` is declared but not referenced in any subsequent `RUN` command — if cache-busting is intentional, it should be used (e.g., `RUN echo $CACHEBUST && ...`), otherwise it can be removed to keep the Dockerfile clean.
- The `secrets` wiring in `docker-compose.yml` looks correct for a standard Docker BuildKit setup. Keep in mind this requires BuildKit to be enabled (which is the default in recent Docker versions).
- If this is deployed via Dokploy or a similar PaaS, make sure the platform supports passing build secrets this way — some managed build environments may not expose `--secret` mounts.

Overall this is a solid improvement to the credential handling in the build pipeline.

---

backend/Dockerfile
lajczi: can I just remove `CACHEBUST` arg? In Dokploy you can use the "Force Rebuild" button which clears docker's build cache achieving the same effect

lajczi: otherwise I can't really use secrets because it just won't work with cache

Norbiros: We don't want to force rebuild of the entire application every time.

Norbiros:
This issue is really complex and probably requires complete rewrite of pulling & providing tasks for the core web server. If you want you can look into that, but first please, discuss it with me.

PR #868, opened 2026-05-01
feat: improve onboarding questions

### LINKED ISSUE #837: [Enhancement]: Improve onboarding/profiling questions
### My idea

Current onboarding dialog looks like this:

Image

It has multiple issues:
- the "czy chcesz dostawać informacje o kolejnych wydarzeniach" question is not really visible
- we should not ask for dietary preferences during this step
- that form has very low completion rate (users are just pressing first option)

Norbiros:
Closes #837 

Before:
image

After:
Screenshot From 2026-05-01 11-54-00
Screenshot From 2026-05-01 11-54-41
Screenshot From 2026-05-01 11-54-52
Screenshot From 2026-05-01 11-55-07


backend/src/routes/account/submit_personal_information.rs
```diff
@@ -39,21 +39,11 @@ pub async fn submit_personal_information(
...
-    if let Some(referral_source) = &request_body.referral_source
-        && referral_source.iter().any(|source| {
-            !ALLOWED_REFERRAL_SOURCES
-                .iter()
-                .any(|allowed_source| allowed_source == source)
-        })
+    if request_body
+        .referral_sources
+        .iter()
+        .any(|source| !ALLOWED_REFERRAL_SOURCES.iter().any(|allowed| allowed == source))
```
Sanielia: make it validation function from actix_web_validation like in 
```rs
#[derive(Debug, Serialize, Deserialize, ToSchema, Validate)]
pub struct CreateTeamModel {
    #[validate(length(min = 3, max = 32), custom(function = "validate_name_chars"))]
    pub team_name: String,
    #[validate(length(min = 3, max = 128), custom(function = "validate_name_chars"))]
    pub organization: Option<String>,
}
```

backend/tests/routes/account/submit_personal_information.rs
```diff
@@ -2,6 +2,50 @@ use crate::test_utils::TestApp;
...
+#[derive(Deserialize)]
+struct PersonalInfoResponse {
+    first_name: String,
+    location: String,
+    ctf_experience: String,
+    school_grade: String,
+    collab_interest: bool,
+    marketing_consent: bool,
+    referral_sources: Option<serde_json::Value>,
+}
```
Sanielia: just use user_personal_info::Model, it is what we return on account/get_personal_information

frontend/app/components/onboarding/StageAbout.vue
```diff
@@ -0,0 +1,151 @@
...
+  <section class="space-y-6">
+    <div>
+      <h2 class="text-2xl font-semibold mb-1">
+        Jak mamy się do Ciebie zwracać?
+      </h2>
+      <p class="text-sm text-muted mb-3">
+        Tak, jak zwracają się do Ciebie znajomi.
+      </p>
+      <UInput
+        v-model="firstName"
```
Sanielia: we need to collect last name somewhere for verifiying people on event, i think it would be best to do it here

frontend/app/components/onboarding/StageConsents.vue
```diff
@@ -0,0 +1,110 @@
...
+          class="text-muted hover:text-primary hover:scale-110 transition duration-200"
+        >
+          <UIcon :name="social.icon" class="w-7 h-7" />
+        </NuxtLink>
+      </div>
+    </div>
+  </section>
+</template>
+
+<style scoped>
```
Sanielia, line 89: why css

frontend/app/components/onboarding/StageWelcome.vue
```diff
@@ -0,0 +1,33 @@
...
+    <p class="text-muted max-w-md">
+      Zanim ruszą zawody, poznajmy się trochę — kilka pytań zajmie chwilę.
+    </p>
+    <p class="text-muted max-w-md text-sm">
+      Twoje odpowiedzi są widoczne wyłącznie dla organizatorów.
+    </p>
+  </section>
+</template>
+
+<style scoped>
```
Sanielia, line 23: why css

frontend/app/pages/account/onboarding.vue
```diff
@@ -1,123 +1,204 @@
...
+        :loading="submitting"
+        @click="onSubmit"
+      >
+        Zapisz i wejdź
+      </UButton>
+    </div>
   </div>
 </template>
+
+<style scoped>
```
Sanielia, line 162: why css

frontend/app/pages/panel/profile.vue
```diff
@@ -47,9 +47,6 @@ async function logout() {
           Ustawienia konta
         </h1>
         <div class="flex flex-col gap-3 mt-3 justify-center">
-          <UButton icon="mdi:account" variant="ghost" color="neutral" @click="navigateTo('/account/submit_personal_info')">
```
Sanielia, line 50: why

Sanielia (changes requested):
unfortunately i couldn check how it looks because of type errors

frontend/app/components/onboarding/StageConsents.vue
Norbiros: There is no other way to specify complex styling for vue transitions than CSS

frontend/app/components/onboarding/StageWelcome.vue
Norbiros: Please, don't spam with your comments.

frontend/app/components/onboarding/StageAbout.vue
Norbiros: As I mentioned in my commit description, these are universal profiling questions and should remain the same regardless of the specific edition.

If we want to collect personal information, GDPR requires a valid reason, defined time frame and the user’s consent. When users provide their details while registering for a specific event, we can simply clear that data after the registration process is complete which is not the case with universal user information (profiling questions).

frontend/app/pages/panel/profile.vue
Norbiros: Why should users be able to modify their answer to “How did you hear about our organization?” That does not really make sense.

Profiling questions should generally be answered once and then remain unchanged. The only exception is the “How should we call you?” question. I did not want to add a separate question flow just for this one field.

I will probably add an option to change that setting in the near future, when redesigning the user panel.

Sanielia: okey, in that case do you want to block it completely or just remove it from panel (in current state users still can just send request on backend)

Sanielia (changes requested):
Image

those headers and placeholder text in input boxes could be smaller on mobile, currently it looks really stuffed

https://github.com/user-attachments/assets/f7ee463c-39ed-4d75-a4e8-28050848eda2

when changing expierience level on mobile, background jumps a lot

https://github.com/user-attachments/assets/50818d63-4706-4901-a393-57f7681b8956

when toggling consents background jumps and size of form changes

what do you think about allowing user to specify referal source when clicking inne?

Sanielia (approved):

PR #874, opened 2026-05-04
fix(frontend): ensure auth fetch retries after token refresh

Norbiros:
Fixes an issue where requests were not retried after a successful token refresh. Adds tests covering this behavior and refactors the shared logic to ensure `$api` and `$auth` handle it consistently.


Sanielia (approved):
lgtm

PR #900, opened 2026-05-13
feat(backend): Add announcements system

### LINKED ISSUE #891: [Feat]: Announcements system
### My suggestion

Create a universal announcement system for organizer-to-participant communication. Announcements should support typed payloads, for example:

* `normal` — a plain text announcement displayed in the panel
* `task_status_update { taskId }` — information about a task status change (e.g. task is down, broken, or fixed)

Ideally, the system should also support:

* fetching the latest task status updates together with the current task state
  (each status update should update the task status and optionally include a comment)
* endpoints for retrieving all task status updates, preferably with pagination

Sanielia:
Resolves: #891


backend/src/models/announcement.rs
```diff
@@ -0,0 +1,180 @@
...
+
+        let mut active_model: announcement::ActiveModel = model.into();
+        active_model.action_type = Set(action_type);
+        active_model.additional_data = Set(action_data);
+
+        let updated_model = active_model.update(database).await?;
+        Ok(updated_model)
+    }
+
+    pub async fn get_latest_task_statuses(
```
Norbiros:
```suggestion
    pub async fn latest_task_statuses(
```

Same for the methods above

backend/src/models/email_verification_request.rs
```diff
@@ -18,8 +18,8 @@ pub enum EmailVerificationAction {
 
 impl EmailVerificationAction {
     pub fn get(&self) -> (String, Option<Value>) {
-        let value = serde_json::to_value(self)
-            .expect("EmailVerificationAction must be serializable");
+        let value =
```
Norbiros, line 21: I think it is unrelated

backend/src/routes/admin/announcements/create.rs
```diff
@@ -0,0 +1,31 @@
...
+    ),
+    operation_id = "admin-announcements-create",
+    tag = "admin/announcements"
+)]
+#[post("/")]
+pub async fn create(
+    app_state: Data<AppState>,
+    action: web::Json<AnnouncementAction>,
+) -> Result<HttpResponse, Error> {
+    let action = action.into_inner();
```
Norbiros, line 24: I didn't look at the full code base, but I think that you shouldn't have logic like invalidating tasks directly in route handler. What if you want to add a new source for loading status updates and you are going to forget about invalidating? 

Ideally this should be in a separate method

Norbiros (changes requested):

backend/src/models/email_verification_request.rs
Sanielia: this looks like linter artifact, i will check it

Sanielia:
```diff
sanielia@sanielia-laptop ~/Documents/Gits/Hack4KrakSite/backend $ bun lint
$ cargo fmt -- --check && cargo clippy
Diff in /home/sanielia/Documents/Gits/Hack4KrakSite/backend/src/models/email_verification_request.rs:18:
 
 impl EmailVerificationAction {
     pub fn get(&self) -> (String, Option<Value>) {
-        let value = serde_json::to_value(self)
-            .expect("EmailVerificationAction must be serializable");                                                                                                                                                               
+        let value =                                                                                                                                                                                                                
+            serde_json::to_value(self).expect("EmailVerificationAction must be serializable");                                                                                                                                     
         let name = value                                                                                                                                                                                                           
             .get("name")
             .and_then(|v| v.as_str())
error: script "lint" exited with code 1
```
yeah, i will create separate commit for that

backend/src/models/announcement.rs
```diff
@@ -0,0 +1,127 @@
...
+        database: &DatabaseConnection,
+        action: &AnnouncementAction,
+    ) -> Result<(), Error> {
+        let (action_type, action_data) = action.get();
+
+        announcement::Entity::insert(announcement::ActiveModel {
+            id: Set(Uuid::new_v4()),
+            action_type: Set(action_type),
+            additional_data: Set(action_data),
+            created_at: Set(Utc::now().naive_utc()),
```
Norbiros, line 71: NP: Do you need to generate this data by hand? Wouldn't it be better to let the db handle it

backend/src/models/announcement.rs
```diff
@@ -0,0 +1,127 @@
...
+        .await?;
+
+        Ok(())
+    }
+
+    pub fn action(&self) -> Result<AnnouncementAction, Error> {
+        let mut value = self
+            .additional_data
+            .clone()
+            .unwrap_or(serde_json::json!({}));
```
Norbiros:
```suggestion
            .unwrap_or_default
```

I am unsure how much of this code is copied from emails, but I see a ton of clones here

backend/src/models/announcement.rs
```diff
@@ -0,0 +1,127 @@
...
+        if let Some(obj) = value.as_object_mut() {
```
Norbiros:
```suggestion
        if let Some(object) = value.as_object_mut() {
```

backend/src/models/announcement.rs
```diff
@@ -0,0 +1,127 @@
...
+            value = serde_json::json!({
+                "type": self.action_type
+            });
+        }
+
+        let action: AnnouncementAction = serde_json::from_value(value)?;
+        Ok(action)
+    }
+
+    pub async fn find_paginated(
```
Norbiros, line 103: In backend rewrite v3 I am going to create universal layer between db models and access, to add proper implementation of pagiantion for all models

backend/src/routes/announcements/task_status_state.rs
```diff
@@ -0,0 +1,20 @@
...
+use actix_web::{HttpResponse, get};
+
+#[utoipa::path(
+    responses(
+        (status = 200, description = "Map of task IDs to their current status.", body = HashMap<String, TaskStatus>),
+    ),
+    tag = "announcements"
+)]
+#[get("/task-status/state")]
+pub async fn task_status_state(app_state: Data<AppState>) -> Result<HttpResponse, Error> {
```
Norbiros: *I hate github, I accidentally pressed cancel and it removed my comment*

1. IMO, we don't care that underneath task statuses are stored as announcements. Those are tasks statuses, so this data should be exposed in the `/tasks` scope
2. This data should be directly integrated into existing endpoints. We are going to always need both statuses and task information (fe. in the list endpoint). Calling 2 separate endpoints (especially in our situation without proper paralization) will be slower, than just always including this data. This also helps us to avoid writing more and more logic in frontend

backend/src/services/task_manager.rs
```diff
@@ -58,7 +63,7 @@ impl TaskManager {
         Self::load_tasks(&self.tasks).await;
     }
 
-    pub fn get_tasks_sorted(&self) -> Vec<SimpleTask> {
+    pub fn tasks(&self, sort: bool) -> Vec<SimpleTask> {
```
Norbiros: I tend to avoid definitions like this. You cannot define defaults, you do not use it as a named argument, so it might be confusing for people without ide (for example on github), to see `tasks(true)`. You can just have 2 methods: `tasks` and `sorted_tasks`. You don't need to duplciate anything

Norbiros (changes requested):

backend/src/models/announcement.rs
Sanielia: if i remeber correctly defaults don't work properly with this orm :/ i will check it tho, if it doesn't then i will create an issue for looking into that and possibly replacing it everywhere

backend/src/models/announcement.rs
Sanielia: i copied issue description to agent so it probably inspired them a bit

backend/src/models/announcement.rs
Norbiros: The issue might be caused by our use of SQLite for testing, but I think timestamps should work properly

Sanielia: unfortunately using ..Default::default() with removal of created_at manual declaration doesnt work :/

PR #912, opened 2026-05-20
fix: Forbid team leader from leaving team

### LINKED ISSUE #911: [Bug]: Orphaned unremovable teams
### Description

Apparently by executing `/teams/membership/leave_team` as a only member (possibly also by unregistering from the event) you can create orphaned teams.

When I tried to remove such team using `/admin/teams/delete/{id}` I got TeamLeaderNotFound error.

### Steps to reproduce

1. Register
2. Leave team using an endpoint above
3.

_No response_

Sanielia:
Resolves: #911

### DIFF
--- a/backend/src/models/teams.rs
+++ b/backend/src/models/teams.rs
@@ -149,22 +149,39 @@ impl teams::Model {
             return Err(Error::Team(UserDoesntBelongToYourTeam));
         }
 
-        if user_to_remove.is_leader {
-            return Err(Error::Team(UserCantRemoveTeamLeader));
+        match Self::remove_user(database, user).await {
+            Err(Error::Team(CannotRemoveTeamLeaderFromTeam)) => {
+                Err(Error::Team(UserCantRemoveTeamLeader))
+            }
+            result => result,
         }
-
-        Self::remove_user(database, user_to_remove).await
     }
 
     pub async fn remove_user(
         database: &DatabaseConnection,
         user: users::Model,
     ) -> Result<(), Error> {
+        if user.is_leader {
+            return Err(Error::Team(CannotRemoveTeamLeaderFromTeam));
+        }
+
         let mut active_user: users::ActiveModel = user.into();
         active_user.team = Set(None);
         active_user.update(database).await?;
 
         Ok(())
+    }
+
+    pub async fn leave_team(
+        database: &DatabaseConnection,
+        user: users::Model,
+    ) -> Result<(), Error> {
+        match Self::remove_user(database, user).await {
+            Err(Error::Team(CannotRemoveTeamLeaderFromTeam)) => {
+                Err(Error::Team(TeamLeaderCannotLeaveTeam))
+            }
+            result => result,
+        }
     }
 
     pub async fn rename(
--- a/backend/locales/error.yaml
+++ b/backend/locales/error.yaml
@@ -71,6 +71,7 @@ TeamError:
   TeamNotFound: "Zespół nie został znaleziony"
   InvalidRegistrationPeriod: "Rejestracja jest niedostępna poza wyznaczonym okresem"
   CannotRegisterInInternalMode: "Nie możesz samemu stworzyć drużyny na to wydarzenie"
+  CannotRemoveTeamLeaderFromTeam: "Nie można usunąć lidera z zespołu."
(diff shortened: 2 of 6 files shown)


backend/src/models/teams.rs
```diff
@@ -160,13 +160,30 @@ impl teams::Model {
...
+        }
+
         let mut active_user: users::ActiveModel = user.into();
         active_user.team = Set(None);
         active_user.update(database).await?;
 
         Ok(())
     }
 
+    pub async fn leave_team(
```
Norbiros, line 175: Is this method unused or am I missing something?

Sanielia: bruh how did i forgot it, it will be used in teams/membership/leave_team route in like 30 seconds

backend/src/models/teams.rs
```diff
@@ -160,13 +160,30 @@ impl teams::Model {
         database: &DatabaseConnection,
         user: users::Model,
     ) -> Result<(), Error> {
+        if user.is_leader {
+            // Theoretically it shouldn't happen, because in all current calls for this method we check it beforehand,
+            // however we want to be sure that in future we don't miss it in any route
```
goteusz-maszyk: Whats the point of this comment its just normal validation

backend/src/models/teams.rs
goteusz-maszyk, line 175: What's the point of adding a separate method doing literally the same

goteusz-maszyk (changes requested):
You still don't cover a situation where for any reason we have a team without leaders, remove team in this case (the same should happen if leader is the only person in a team)

Sanielia:
> You still don't cover a situation where for any reason we have a team without leaders, remove team in this case (the same should happen if leader is the only person in a team)

I didn't cover it because it's expected behavior, we expect teams to have no leaders in external registration mode, because authority that adds people to teams is school/organization not participants, creating separate table for external teams is an overkill and we can assure there is no team without leader in internal registration by making team creator a leader and blocking removal of team leader from team

as per removing team in case it vconsist only of its leader, why would we do that? if we allow teams with 1 participants and treat them as a team, because team is what participates in event, submits flags and get points, not participant or group of them and there is no need to create separation between 1 participant or multiple participant teams, because why would there be

backend/src/models/teams.rs
Sanielia: point is that one day someone might think, why don't we just delete this method if we check it anyway in other methods, and then remove it and during code review it could be that no one notices it, this comment is here to make those people think before removing it or accepting removal

backend/src/models/teams.rs
Sanielia: boilerplate can be useful on architectural level
we want to assure that team leader is never removed from the team, therefore we use one commont method for removing user from a team which is remove_user and it check for user being a leader

at the same time we also want to return different errors depending on the method, if user wants to leave the team we want to say you cannot do that as a team leader, if user wants to remove team leader we say, no you cannot remove team leader, also different error code on expected path with it just being forbidden and unexpected path which shouldn't happen with server error

it could probably be improved by:
1. changing how errors are turned into http responses by introducing error variants to which i say too much work, too much not the scope of this pr, create issue for that
2. changing how error is propagated and just adding error handling to public methods (i also need to make remove_user private before merging this pr) instead of duplicating the is_leader check, which i might do but still separate methods will stay

goteusz-maszyk (dismissed):
lgtm ship it :rocket: :rocket:
> handle_remove_user_from_team_error
> tellWitnessesThatIWasMurdered

backend/src/utils/error_handling.rs
```diff
@@ -14,3 +15,13 @@ where
...
+
+pub fn handle_remove_user_from_team_error(
+    result: Result<(), Error>,
+    returned_error: Error,
+) -> Result<(), Error> {
+    match result {
+        Err(Error::Team(CannotRemoveTeamLeaderFromTeam)) => Err(returned_error),
+        result => result,
+    }
+}
```
Norbiros: I can't see why did you add this function. Either make it a universal utility, that allows for handling any single error or just duplicate those 3 lines of code

Norbiros (approved):

